PR #8, opened 2019-04-17
fix(colors): add 03 gray

### LINKED ISSUE #2: GLOBAL > Grey color change
Change light grey #FAFAFA to #**F7F7F7**

adamraider:
fix #2

- removes $ray-color-gray-02


cmugla:
Wondering if this is a system color change or a new color, specifically because 02 and 03 are so similar.... if it is a system change, we may want to remove `-02` and have it be a breaking change, so people can't use that color anymore. Thoughts?

adamraider:
Will need to remove -02 from the system. This will be a breaking change.

PR #10, opened 2019-04-18
feat(chip): add the chip component

adamraider:
image

### DIFF
--- a/packages/core/src/components/chip/index.js
+++ b/packages/core/src/components/chip/index.js
@@ -2,7 +2,9 @@ class Chip {
   static instances = new WeakMap();
 
   static get cssClasses() {
-    return {};
+    return {
+      ACTIVE: 'ray-chip--active'
+    };
   }
 
   static get strings() {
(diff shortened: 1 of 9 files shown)


packages/core/src/components/chip/index.js
```diff
@@ -0,0 +1,68 @@
+class Chip {
```
adamraider: This needs tests before we merge.

packages/core/src/components/chip/index.js
```diff
@@ -0,0 +1,68 @@
...
+    this.state = {
+      active: options.active
+    };
+
+    this._bindEventListeners();
+    this.assignClasses();
+  }
+
+  _bindEventListeners() {
+    this._root.addEventListener('mousedown', this.onMousedown);
```
cmugla: Thoughts on adding a `touchstart` event as well.. not sure what the best UX is for this - seeing on my phone that they don't work very well for quicker, slightly clumsier gestures

packages/core/src/components/chip/index.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  assignClasses() {
+    if (this.state.active) {
+      this._root.classList.add('ray-chip--active');
+    } else {
+      this._root.classList.remove('ray-chip--active');
+    }
+  }
```
maniator: Can't this be done with plain css?

adamraider: I don't want to clutter the DOM with checkbox elements

adamraider:
Closing in favor of #181

PR #16, opened 2019-04-26
feat(textarea): allow multi-row textarea

i8ramin:
when rows are provided to a textarea, the fixed height is removed and set to "auto"

Screen Shot 2019-04-26 at 2 28 38 PM

### DIFF
--- a/packages/core/src/components/text-field/_text-field.scss
+++ b/packages/core/src/components/text-field/_text-field.scss
@@ -170,7 +170,7 @@
     }
   }
 
-  textarea[rows].#{$ray-class-prefix}text-area {
+  [rows].#{$ray-class-prefix}text-area {
     &__input {
       height: auto !important;
     }
--- a/packages/core/src/components/text-field/_text-field.scss
+++ b/packages/core/src/components/text-field/_text-field.scss
@@ -172,7 +172,8 @@
 
   [rows].#{$ray-class-prefix}text-area {
     &__input {
-      height: auto !important;
+      height: auto; // ie11 fallback
+      height: unset;
     }
   }
 }
--- a/packages/core/src/components/text-field/_text-field.scss
+++ b/packages/core/src/components/text-field/_text-field.scss
@@ -169,4 +169,10 @@
       }
     }
   }
+
+  textarea[rows].#{$ray-class-prefix}text-area {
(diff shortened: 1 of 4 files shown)


packages/core/src/components/text-field/_text-field.scss
```diff
@@ -169,4 +169,10 @@
       }
     }
   }
+
+  textarea[rows].#{$ray-class-prefix}text-area {
```
adamraider: Is the `textarea` part of this selector necessary?

adamraider:
@i8ramin this is cool, could you add a story for it in storybook? So we can visually track it with chromatic.

packages/core/src/components/text-field/_text-field.scss
i8ramin: probably not. could do `*[rows]` .. but `textarea` is only element that takes in `rows` and none of this really applies to any other html element

i8ramin:
> @i8ramin this is cool, could you add a story for it in storybook? So we can visually track it with chromatic.

There was actually already a storybook. It was just using the wrong attribute -- https://github.com/wework/ray/pull/16/commits/881e1d2b8b79a08051faac9866435bb575564cfe#diff-9008c83448da3b291989f05165b793edR68

packages/core/stories/text-field.stories.js
```diff
@@ -65,7 +65,7 @@ storiesOf('Text Field', module)
       <div className="ray-text-area">
         <textarea
           className="ray-text-area__input"
-          row="4"
+          rows="4"
```
adamraider: nice

packages/core/src/components/text-field/_text-field.scss
```diff
@@ -169,4 +169,10 @@
       }
     }
   }
+
+  [rows].#{$ray-class-prefix}text-area {
+    &__input {
+      height: auto !important;
```
adamraider:
```suggestion
      height: auto; // ie11 fallback
      height: unset;
```

`auto` value prevents the user from resizing. Let's use unset and leave auto as a fallback for ie.

adamraider:
@allcontributors[bot] please add @i8ramin for code

Issue #81, opened 2019-05-09
Phone + Number fields used on vanilla wework.com


i8ramin:
![Screen Shot 2019-05-09 at 11 05 05 AM](https://user-images.githubusercontent.com/49030/57464462-c6cd4b00-724a-11e9-8596-74f07d1de935.png)

Would be nice to add the Phone and Number Selector inputs that are currently used on wework.com to Ray. I know that the Phone one isn't the best implementation and there is room for improvement, but the lack of any support in Ray right now leaves it up to each team to copy/paste whats there vs consolidating it all in one place (so we can easily replace it later)

adamraider:
Can we close this?

PR #82, opened 2019-05-09
feat(select): prepend support on textfield & select

i8ramin:
No longer breaking change. Decided to not touch the current `with-icon` implementation and create new styles to support `prepend`.

## Main motivation (re #81)

148bf077b51e186b763e56e7423305df _Screen%20Shot%202019-05-09%20at%2011 05 05%20AM

### Screenshots

Screen Shot 2019-05-10 at 10 31 16 AM

Screen Shot 2019-05-09 at 2 02 06 PM

Screen Shot 2019-05-09 at 2 02 52 PM

### DIFF
--- a/packages/core/src/components/select/_select.scss
+++ b/packages/core/src/components/select/_select.scss
@@ -5,6 +5,7 @@
   .#{$ray-class-prefix}select {
     @include required-support('select');
     @include icon-support('select');
+    @include prepend-support('select');
 
     border: $ray-border-width solid $ray-color-gray-60;
     background-color: $ray-color-white;
@@ -12,6 +13,7 @@
     position: relative;
     height: $ray-field-height;
     display: flex;
+    align-items: stretch;
 
     // __icon--right is DEPRECATED in favor of __icon--end, and will be
     // removed in a future release of Ray. Use of __icon-right may result
@@ -20,11 +22,11 @@
     &__icon--end {
       @include icon;
       // need to add spacing to compensate for the arrow
-      right: calc(#{$ray-field-h-spacing} * 2 + 12px);
+      right: calc((#{$ray-field-h-spacing} * 2) + #{$ray-field-h-spacing});
 
       [dir='rtl'] & {
         right: 0;
-        left: calc(#{$ray-field-h-spacing} * 2 + 12px);
+        left: calc((#{$ray-field-h-spacing} * 2) + #{$ray-field-h-spacing});
       }
     }
 
@@ -44,9 +46,13 @@
       border-radius: 2px;
 
       [dir='rtl'] & {
-        position: unset;
-        margin-left: $ray-field-h-spacing;
+        right: calc(100% - (2 * #{$ray-field-h-spacing}));
       }
+    }
+
+    &__wrapper {
+      position: relative;
+      flex: 1;
     }
 
     &__label {
@@ -162,6 +168,28 @@
 
       .#{$ray-class-prefix}select__label {
         color: rgba($ray-color-gray-60, 0.4);
+      }
+    }
+
+    &--simple {
+      flex: 1;
+      border: none;
+      background-color: transparent;
+      border-radius: 0;
+
+      &__input {
+        font-size: ($ray-field-label-size * 1.2);
+        width: rem(69px);
+        padding-right: 3 * $ray-field-h-spacing;
+
+        [dir='rtl'] & {
+          padding-right: $ray-field-h-spacing;
+          padding-left: 3 * $ray-field-h-spacing;
+        }
+
+        &::-ms-expand {
+          display: none;
+        }
       }
     }
   }
--- a/packages/core/src/components/text-field/_text-field.scss
+++ b/packages/core/src/components/text-field/_text-field.scss
@@ -5,6 +5,7 @@
   .#{$ray-class-prefix}text-field {
     @include required-support('text-field');
     @include icon-support('text-field');
+    @include prepend-support('text-field');
   }
 
   .#{$ray-class-prefix}text-area {
@@ -19,6 +20,13 @@
     border-radius: $ray-border-radius;
     position: relative;
     height: $ray-field-height;
+    display: flex;
+    align-items: stretch;
+
+    &__wrapper {
+      position: relative;
+      flex: 1;
+    }
 
     &__label {
       @include label;
--- a/packages/core/src/global/mixins/_form-items.scss
+++ b/packages/core/src/global/mixins/_form-items.scss
@@ -219,6 +219,18 @@
     }
   }
 
+  // --icon-left/right are DEPRECATED in favor of --icon-start/end,
+  // and will be removed in a future release of Ray. Use of --icon-left/right may
+  // result in confusing behavior for RTL language sensitive layouts.
+  &--disabled {
+    .#{$ray-class-prefix}#{$class}__icon--left,
+    .#{$ray-class-prefix}#{$class}__icon--right,
+    .#{$ray-class-prefix}#{$class}__icon--start,
+    .#{$ray-class-prefix}#{$class}__icon--end {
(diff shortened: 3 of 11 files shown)


adamraider:
@i8ramin Interesting contribution, however, this will need to be audited by design. Can you please confer with Dana Ballasy about this change?

adamraider:
What is the breaking change for?

i8ramin:
> What is the breaking change for?

@adamraider, there are new wrapping div's that need to be added for the with-icon and with-prepend stuff to work. I'll point them out in the PR

cmugla:
hmmm.. so the idea is you would prepend a select to the text field in order to solve for phone field? 

Might be useful to provide an example of this specific use case, since that is the reason for adding? like `textfield with a select` story

docs/html/select/select-with-icon.html
```diff
@@ -1,22 +1,25 @@
...
-      <option value="Pikachu">WeWork</option>
-      <option value="Squirtle">WeLive</option>
-      <option value="Bulbasaur">WeGrow</option>
-      <option value="Charmander">WeTech</option>
-    </select>
-
-    <label class="ray-select__label">
-      Better Together
-    </label>
+    <div class="ray-select__icon--left">
```
i8ramin: BREAKING CHANGE: the (svg) icon no longer receives the `ray-select__icon--left` class. There is now a wrapping div around the svg icon.

docs/html/select/select-with-icon.html
```diff
@@ -1,22 +1,25 @@
...
+      <svg viewBox="0 0 25 25" style="height: 1rem;">
+        <g id="budicon-profile-picture">
+          <path
+            d="M12.5,4A4.5,4.5,0,1,0,17,8.5,4.5,4.5,0,0,0,12.5,4Zm0,8A3.5,3.5,0,1,1,16,8.5,3.504,3.504,0,0,1,12.5,12Zm0-12A12.4886,12.4886,0,0,0,5.0007,22.4834v0a12.4325,12.4325,0,0,0,14.9983,0v0q.5-.3761.9593-.7988l0,0A12.4869,12.4869,0,0,0,12.5,0Zm0,24a11.4432,11.4432,0,0,1-7.3931-2.7041,7.4887,7.4887,0,0,1,14.7863,0A11.4432,11.4432,0,0,1,12.5,24Zm8.25-3.5061a8.4871,8.4871,0,0,0-16.5,0,11.5,11.5,0,1,1,16.5,0Z"
+          />
+        </g>
+      </svg>
+    </div>
+    <div class="ray-select__wrapper">
```
i8ramin: BREAKING CHANGE: There is now a "wrapper" that wraps the input + label so that we can position them easier in relation to the icons

packages/core/src/components/select/_select.scss
```diff
@@ -34,6 +36,11 @@
       border-radius: 2px;
     }
 
+    &__wrapper {
```
i8ramin: see https://github.com/wework/ray/pull/82/files#r282620989

packages/core/src/global/mixins/_form-items.scss
```diff
@@ -150,26 +150,34 @@
...
-  // optical correction, subtract 1px
-  top: $ray-field-v-spacing - rem(1px);
 
   path {
     fill: currentColor;
   }
 }
 @mixin icon-support($class) {
   &--with-icon-left {
+    &.#{$ray-class-prefix}#{$class}--active,
```
i8ramin: Some adjustments for the label and input to appear correctly when the input is active or has value

packages/core/stories/text-field.stories.js
```diff
@@ -320,49 +320,118 @@ storiesOf('Text Field', module)
...
-              id="input2"
-              placeholder="Few people are aware..."
-              disabled
-            />
-            <label className="ray-text-field__label" htmlFor="input2">
-              Fun fact about Ray Eames
-            </label>
+    return (
+      <>
+        <div className="ray-form-item">
```
cmugla: what is the difference between these two form items? possible copy/paste oop?
Screen Shot 2019-05-09 at 3 10 57 PM

cmugla:
doesn't look like disabled behaves the same way with `-prepend`
Screen Recording 2019-05-09 at 03 14 PM

packages/core/stories/text-field.stories.js
cmugla: Think the second one is intended for a disabled version (like text field with icon above it)

i8ramin: yeah copy/paste job. ill make the 2nd one look different, or just remove it

cmugla: can use it for disabled state

i8ramin:
> doesn't look like disabled behaves the same way with `-prepend`
> Screen Recording 2019-05-09 at 03 14 PM

That was a bad copy/paste job. It was missing the `disabled` prop. Fixed now

i8ramin:
> hmmm.. so the idea is you would prepend a select to the text field in order to solve for phone field?
> 
> Might be useful to provide an example of this specific use case (with a dummy select), since that is the reason for adding? like `textfield with a select` story

This is just the first step towards that end goal. Adding a phone select in there is gonna be tricky, so didn't want to create a huge PR. This one just makes it so we can "prepend" things to textfield & selects

i8ramin:
@cmugla added a simple flag dropdown (no functionality, just visuals)

Screen Shot 2019-05-10 at 10 31 16 AM

packages/core/src/components/select/_select.scss
```diff
@@ -140,5 +147,22 @@
         color: rgba($ray-color-gray-60, 0.4);
       }
     }
+
+    &--simple {
```
i8ramin: new addition to `ray-select` to support things like the flag dropdown

Screen Shot 2019-05-10 at 10 34 27 AM

packages/core/src/components/select/_select.scss
```diff
@@ -140,5 +147,22 @@
...
+      flex: 1;
+      border: none;
+      background-color: transparent;
+      border-radius: 0;
+
+      &__input {
+        font-size: ($ray-field-label-size * 1.2);
```
i8ramin: this is specifically to make the flags a but larger .. not sure if we should set this font size here

packages/core/src/components/text-field/_text-field.scss
```diff
@@ -3,7 +3,9 @@
 
 @include exports('ray-text-field') {
   .#{$ray-class-prefix}text-field {
+    @include required-support('text-field');
```
i8ramin: noticed Ray input labels didnt really have support for adding (*) to required fields

adamraider:
@i8ramin Is there any way to avoid the breaking change? We _just_ released 1.0.0 lol

packages/core/src/components/text-field/_text-field.scss
cmugla: Cool!

i8ramin:
> @i8ramin Is there any way to avoid the breaking change? We _just_ released 1.0.0 lol

@adamraider its only breaking if you used `with-icon` support in textfield and selects. otherwise, everything works as expected.

packages/core/src/components/select/_select.scss
```diff
@@ -44,11 +46,15 @@
       border-radius: 2px;
 
       [dir='rtl'] & {
-        position: unset;
-        margin-left: $ray-field-h-spacing;
+        right: calc(100% - (2 * #{$ray-field-h-spacing}));
```
i8ramin: This is a slight change in how we handle the arrow on RTL. Instead of unsetting position, we calculate it so that the behavior is similar to how it is done with LTR

i8ramin:
Screen Shot 2019-05-28 at 3 59 35 PM

PR #87, opened 2019-05-10
fix: add required support for labels

i8ramin:
for required text-field and selects, the label will automatically get an asterik *

### DIFF
--- a/packages/core/src/global/mixins/_form-items.scss
+++ b/packages/core/src/global/mixins/_form-items.scss
@@ -230,3 +230,14 @@
     }
   }
 }
+
+@mixin required-support($class) {
+  &--required {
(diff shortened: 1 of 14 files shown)


packages/core/src/global/mixins/_form-items.scss
```diff
@@ -230,3 +230,14 @@
     }
   }
 }
+
+@mixin required-support($class) {
+  &--required {
+    .#{$ray-class-prefix}#{$class}__label {
+      &::after {
+        content: '*';
+        padding-left: 3px;
```
adamraider: Is this value arbitrary?

i8ramin: i guess, i took it from vanilla www

PR #89, opened 2019-05-12
fix(select): don't move label on focus/active

### LINKED ISSUE #88: Select - label placement on active
When a select element has focus, the label moves to the top of the field (like other inputs), however if no selection is made (user clicks outside of the dropdown), the label remains there until focus is lost (blur). Unfortunately, the browser does not trigger _blur_ until another user click (or selection) happens. This can be seen below:

Screen Recording 2019-05-12 at 10 14 AM

i8ramin:
fix #88

The current Select focus/active implementation moves the label up when a select is activated, but if no selection is made (and user clicks off of the select), then the label remains "activated" and only returns back to normal when the select completely loses focus. This change makes it so that the label only moves up if a selection is made.

## Current behavior

Screen Recording 2019-05-12 at 10 24 AM

## New proposed behavior

Screen Recording 2019-05-12 at 10 26 AM

### DIFF
--- a/packages/core/src/components/select/_select.scss
+++ b/packages/core/src/components/select/_select.scss
@@ -88,19 +88,8 @@
     &--active {
       border-color: $ray-color-blue-50;
 
-      .#{$ray-class-prefix}select__input {
-        color: $ray-color-black;
-      }
-
       .#{$ray-class-prefix}select__label {
-        @include label--active;
         color: $ray-color-blue-50;
-      }
-
-      &.#{$ray-class-prefix}select--placeholder-mode {
-        .#{$ray-class-prefix}select__input {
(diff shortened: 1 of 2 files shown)


i8ramin:
> This breaks Chrome+Windows

Right you are. Fixed https://github.com/wework/ray/pull/89/commits/832de7ce3f5ee6852cc9afdd27407ebb7384d2d6

i8ramin:
> This breaks Chrome+Windows

it sure does. that placeholder does complicate things. do we really need it?

i8ramin:
Attempt at fixing placeholder mode across browsers -- https://github.com/wework/ray/pull/89/commits/5aa321b0e7d39f714647a877d01c843d7cf392c3

PR #100, opened 2019-05-16
fix(tag): add link styles

cmugla:
Link styles:
Screen Recording 2019-05-16 at 06 59 PM

reg styles:
Screen Shot 2019-05-16 at 7 00 50 PM


adamraider:
Nice, can you add a story so we can track it with Chromatic?

PR #101, opened 2019-05-16
feat(card): add ray-card--row

cmugla:
Screen Recording 2019-05-17 at 01 41 PM

TODO:
- [x] stack on mobile/tablet
- [x] add in the heading font styles 
- [x] add all the aspect ratios


packages/core/src/components/card/_card.scss
```diff
@@ -40,20 +40,42 @@
...
+      flex-direction: row;
+
+      .#{$ray-class-prefix}card__image,
+      .#{$ray-class-prefix}card__content {
+        width: 50%;
+        float: left;
+      }
+
+      .#{$ray-class-prefix}image--16by9 {
+        padding-bottom: calc(50% / (16 / 9));
```
cmugla: these should probably be dynamically generated with out supported ratios...

cmugla:
we may want to add breakpoint logic to this

packages/core/src/components/card/_card.scss
```diff
@@ -40,20 +40,42 @@
       }
     }
 
+    &--row {
+      flex-direction: row;
+
+      .#{$ray-class-prefix}card__image,
+      .#{$ray-class-prefix}card__content {
+        width: 50%;
+        float: left;
```
i8ramin: Why not use flexbox?

docs/html/card/card--row.html
```diff
@@ -0,0 +1,16 @@
+<div class="ray-card ray-card--row" style="max-width: 400px;">
+  <div class="ray-card__image ray-image ray-image--16by9">
```
adamraider: Since this container won't conform to any aspect ratio (due to the fixed height/variable width), we'll probably want to use a bg here

packages/core/src/components/card/_card.scss
cmugla: oh right, duh!

packages/core/src/components/card/_card.scss
cmugla: done!

docs/html/card/card--row.html
cmugla: added!

packages/core/stories/card.stories.js
```diff
@@ -2,8 +2,14 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 
 import getPlaceholderURL from './util/placeholder';
+import settings from '../src/global/js/settings';
+
+const viewports = Object.values(settings.breakpointsInPixels).filter(Boolean);
```
adamraider: we'll probs want to put this in a util, but this is cool for now

PR #106, opened 2019-05-22
feat: support right-to-left languages

iainmcgin:
This makes the Ray component styles sensitive to direction, such that components will render in the
expected orientation for right-to-left languages. Ideally, this would have been done using CSS' new
Logical Properties, but support for these is not pervasive yet, so style overrides are applied using
parent selectors wherever a left-right asymmetrical style is applied. The examples have been
extended to show every component with Hebrew right-to-left rendering.

BREAKING CHANGE: Component modifiers that ended in "left" or "right" have been changed to "start"
and "end" respectively, to communicate the alignment relative to the expected flow of text in the
host language. All examples have been updated, but any clients using Ray will likely need to be
updated. Alternatively, we could temporarily alias make "left" == "start" and "right" == "end"
semantically, until we can fully deprecate those styles.

### DIFF
--- a/packages/core/src/components/breadcrumb/_breadcrumb.scss
+++ b/packages/core/src/components/breadcrumb/_breadcrumb.scss
@@ -20,13 +20,24 @@
     &:not(:last-child) {
       margin-right: 0.5rem;
 
+      [dir='rtl'] & {
+        margin-right: 0;
+        margin-left: 0.5rem;
+      }
+
       &::after {
         margin-left: 0.5rem;
-        content: '➝';
+        content: '→';
         font-size: 1.25rem;
         position: relative;
         top: -6px;
         line-height: 1.5;
+
+        [dir='rtl'] & {
+          margin-right: 0.5rem;
+          margin-left: 0;
+          content: '←';
+        }
       }
     }
 
--- a/packages/core/src/components/button/_button.scss
+++ b/packages/core/src/components/button/_button.scss
@@ -36,6 +36,10 @@
       }
     }
 
+    // --pull-left is DEPRECATED in favor of --pull-start, and will be
+    // removed in a future release of Ray. Use of pull-left may result
+    // in confusing behavior for RTL language sensitive layouts.
+    &--pull-left,
     &--pull-start {
       margin-left: -($ray-button-h-spacing + $ray-border-width);
 
@@ -45,6 +49,10 @@
       }
     }
 
+    // --pull-right is DEPRECATED in favor of --pull-end, and will be
+    // removed in a future release of Ray. Use of pull-right may result
+    // in confusing behavior for RTL language sensitive layouts.
+    &--pull-right,
     &--pull-end {
       margin-right: -($ray-button-h-spacing + $ray-border-width);
 
@@ -156,6 +164,10 @@
     line-height: 1;
     height: auto;
 
+    // button--pull-left is DEPRECATED in favor of button--pull-start, and will
+    // be removed in a future release of Ray. Use of button--pull-left may
+    // result in confusing behavior for RTL language sensitive layouts.
+    &.#{$ray-class-prefix}button--pull-left,
     &.#{$ray-class-prefix}button--pull-start {
       margin-left: -($ray-button-h-spacing-compact + $ray-border-width);
 
@@ -165,6 +177,10 @@
       }
     }
 
+    // button--pull-right is DEPRECATED in favor of button--pull-end, and will
+    // be removed in a future release of Ray. Use of button--pull-right may
+    // result in confusing behavior for RTL language sensitive layouts.
+    &.#{$ray-class-prefix}button--pull-right,
     &.#{$ray-class-prefix}button--pull-end {
       margin-right: -($ray-button-h-spacing-compact + $ray-border-width);
 
--- a/packages/core/src/components/button/_button.scss
+++ b/packages/core/src/components/button/_button.scss
@@ -29,14 +29,29 @@
 
     &:not(:last-of-type) {
       margin-right: 1rem;
+
+      [dir='rtl'] & {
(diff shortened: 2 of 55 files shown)


packages/core/src/components/text-field/_text-field.scss
```diff
@@ -123,12 +123,12 @@
 
       .#{$ray-class-prefix}text-field__label,
       .#{$ray-class-prefix}text-area__label {
-        padding-top: $ray-field-v-spacing-compact;
+        padding-block-start: $ray-field-v-spacing-compact;
```
cmugla: this isn't supported in IE or Edge:

Screen Shot 2019-05-22 at 7 01 12 PM

maniator: It is however supported in the edge canary ;-)

adamraider:
Looks like CI did not run, could you try pushing an empty commit?

It also might be because you're trying to merge from a forked repo. Not sure

adamraider:
> Alternatively, we could temporarily alias make "left" == "start" and "right" == "end"
semantically, until we can fully deprecate those styles.

I would like to do this :)

packages/core/src/components/text-field/_text-field.scss
i8ramin: Does `padding-top` have an impact on right-to-left languages?

i8ramin:
This project is pretty old, but found in some initial research on the topic a while back -- https://github.com/twitter/css-flip. Not sure if helpful in anyway, but thought I'd share

packages/core/src/components/breadcrumb/_breadcrumb.scss
```diff
@@ -20,13 +20,24 @@
...
 
+      [dir='rtl'] & {
+        margin-right: 0;
+        margin-left: 0.5rem;
+      }
+
       &::after {
         margin-left: 0.5rem;
-        content: '➝';
+        content: '→';
```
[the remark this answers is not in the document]
iainmcgin: Unfortunately there is no flipped variant of this arrow in the Dingbats unicode range: The character prior to this change was U+279D "Triangle-Headed Rightwards Arrow". The new character is U+2192 "Rightwards Arrow", which comes from the "Arrows" Unicode block (https://www.compart.com/en/unicode/block/U+2190) that has the equivalent U+2190 "Leftwards Arrow".

Are you concerned about a specific stylistic difference in the arrows? And if so, are you OK with the left arrow for RTL looking different from the right arrow for LTR?

packages/core/src/components/image/_image.scss
```diff
@@ -53,6 +53,10 @@ $ray-image-ratio-map: (
   -webkit-line-clamp: 2;
   -webkit-box-orient: vertical;
   margin-top: $ray-spacing-xs;
+
+  [dir='rtl'] & {
+    text-align: left;
```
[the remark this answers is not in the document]
iainmcgin: I was just mirroring the choice to right-align text in captions in LTR, meaning we would left-align text in captions in RTL. Quite a mental gymnastic workout we're getting here, at least we don't _also_ have to deal with top-to-bottom, right-to-left traditional Japanese script 😆

packages/core/src/components/text-field/_text-field.scss
iainmcgin: Ah my bad, this is a remnant from my first attempt to do what was needed using logical properties. The future "correct" / pedantic way to do everything in future once logical properties are standardized (they're still draft right now) will be to drop all ideas of top / bottom / left / right in favor of block-start / block-end / inline-start / inline-end.

Anyway, it doesn't impact just supporting RTL. I'll revert this.

iainmcgin:
> > Alternatively, we could temporarily alias make "left" == "start" and "right" == "end"
> > semantically, until we can fully deprecate those styles.
> 
> I would like to do this :)

I've reintroduced the left and right labeled rules as aliases of start and end, PTAL.

packages/core/src/components/breadcrumb/_breadcrumb.scss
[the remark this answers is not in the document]
cmugla: Yes, stylistic difference. You have to flip it using css transform

packages/core/src/components/image/_image.scss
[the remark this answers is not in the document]
cmugla: ha, yet! yeah makes sense, good 👁

adamraider:
@iainmcgin Can you try pushing your branch directly to the repo? For some reason CI isn't running.

packages/core/src/components/text-field/_text-field.scss
cmugla: TIL!

packages/core/src/components/button/_button.scss
```diff
@@ -29,14 +29,37 @@
...
+      [dir='rtl'] & {
+        margin-right: 0;
+        margin-left: 1rem;
+      }
     }
 
-    &--pull-left {
+    // --pull-left is DEPRECATED in favor of --pull-start, and will be
+    // removed in a future release of Ray. Use of pull-left may result
+    // in confusing behavior for RTL language sensitive layouts.
```
cmugla: 👏

cmugla: love these comments, thank you

adamraider: This class is undocumented and I plan on removing it all together anyway lol

iainmcgin:
Moved to #110, closing this.

packages/core/src/components/breadcrumb/_breadcrumb.scss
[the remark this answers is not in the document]
iainmcgin: Sneaky 😆 applying this in #110

PR #108, opened 2019-05-23
fix: updates line-height

adamraider:


packages/core/src/global/_variables.scss
```diff
@@ -8,7 +8,7 @@ $ray-font-stack-sans-serif: 'Apercu', Avenir, -apple-system, BlinkMacSystemFont,
 $ray-font-stack-mono: 'Apercu Mono', SFMono-Regular, Menlo, Monaco, Consolas,
   'Liberation Mono', 'Courier New', monospace;
 $ray-base-font-size: rem(16px);
-$ray-base-line-height: 1.75rem;
+$ray-base-line-height: 1.618;
```
alanfriedman: Why change the global line height?

adamraider: it's necessary

adamraider: The global line height should be relative to the element not the base font-size

alanfriedman: So this is an unrelated fix?

adamraider: it's somewhat related

PR #110, opened 2019-05-23
feat: support right-to-left languages

iainmcgin:
[supercedes #106]

This makes the Ray component styles sensitive to direction, such that components will render in the
expected orientation for right-to-left languages. Ideally, this would have been done using CSS' new
Logical Properties, but support for these is not pervasive yet, so style overrides are applied using
parent selectors wherever a left-right asymmetrical style is applied. The examples have been
extended to show every component with Hebrew right-to-left rendering.


adamraider:
The card changes in Chromatic can be fixed by rebasing off master

adamraider:
And the breadcrumb change could be resolved by following @cmugla's suggestion here https://github.com/wework/ray/pull/106#discussion_r287141580.

Other than that, this looks amazing.

PR #134, opened 2019-05-30
chore: remove commitlint

adamraider:
Contributors should not be expected to learn this convention. It should be up to the ray-maintainers, to enforce this practice (and write proper descriptions) when `squash & merging`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,16 +24,6 @@ $ npx lerna run storybook --scope="@wework/ray-core" --parallel
 $ npx lerna run dev --scope="ray-website-gatsby"  --parallel
 ```
 
-## Contributing
-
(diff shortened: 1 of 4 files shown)


i8ramin:
I actually find the commit lint pretty handy. whats the problem we are facing with this? it seems pretty configurable.

adamraider:
Problem is:

1. People are confused about how this convention works and what's its used for (versioning, changelog). I don't want to add friction to the process of contributing for non-maintainers.
2. The convention only matters on the _squashed_ commit (which only a maintainer can make) the commits that are made by the contributor in the pr dont particularly matter.

cmugla:
Please update the docs to include which versioning actions are fired per commitzen keyword

PR #141, opened 2019-06-04
feat(accessibility): only show active states when keyboard navigating

adamraider:
Adds functionality that disables focus states and hover states for buttons when theyre not applicable. This makes the ux a little cleaner and is inline with peoples expectations for these states

## After

image
 :^)

### DIFF
--- a/docs/foundations/accessibility.md
+++ b/docs/foundations/accessibility.md
@@ -17,8 +17,6 @@ If the user is using a keyboard to navigate, Ray will apply a class to the docum
 
 The [Web Content Accessibility Guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG) state that to ensure readability, text and interactive elements should maintain a color contrast ratio of at least 4.5:1.
 
-[Web Content Accessibility Guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG)
-
 ### Color Blindness
 
 The most common variety of colorblindness is red-green color blindness. It's important to **avoid** juxtaposing these colors.
--- a/docs/foundations/accessibility.md
+++ b/docs/foundations/accessibility.md
@@ -21,7 +21,7 @@ The WCAG state that to ensure readability, text and interactive elements should 
 
 ### Color Blindness
 
-The most common variety of colorblindness is red-green color blindness. It's important to **avoid** juxtaposing these color.
+The most common variety of colorblindness is red-green color blindness. It's important to **avoid** juxtaposing these colors.
 
 [US Government Standards](https://designsystem.digital.gov/design-tokens/color/overview/)
 
--- a/docs/foundations/accessibility.md
+++ b/docs/foundations/accessibility.md
@@ -15,7 +15,7 @@ If the user is using a keyboard to navigate, Ray will apply a class to the docum
 
 ### Color Ratio
 
-The WCAG state that to ensure readability, text and interactive elements should maintain a color contrast ratio of at least 4.5:1.
+The [Web Content Accessibility Guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG) state that to ensure readability, text and interactive elements should maintain a color contrast ratio of at least 4.5:1.
 
 [Web Content Accessibility Guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG)
 
--- a/docs/foundations/accessibility.md
+++ b/docs/foundations/accessibility.md
@@ -7,7 +7,7 @@ title: Accessibility
 
 ### Keyboard Navigation
 
-Many people navigate the web using technologies such as screen readers, or are unable to use a traditional mouse. It's imperative that content and interactive elements are accessible by way of keyboard navigation, and clearly indicate when
+Many people navigate the web using technologies such as screen readers, or are unable to use a traditional mouse. It's imperative that content and interactive elements are accessible by way of keyboard navigation, and clearly indicate where the user has focus.
 
 Do not remove focus states from elements unless you implement a suitable replacement.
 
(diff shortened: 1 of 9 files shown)


packages/core/src/global/js/accessibility.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable no-inner-declarations */
+
+function setupAccessibility() {
+  if (typeof document !== 'undefined') {
+    document.addEventListener('mousedown', removeAccessibilityClass);
```
gmharrison: this might warrant more research, but i'm not sure that using a mouse negates the need for accessibility styles. users might use a mouse but still require clear accessible focus states.

adamraider: I analyzed some pretty high profile sites and this was a very common pattern I found

kconst: We should do some future investigation into pointer event support.

packages/core/src/global/js/accessibility.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable no-inner-declarations */
```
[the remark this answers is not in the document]
designjockey: I believe the below pattern will resolve this:


```
const ACCESSIBILITY_CLASS = 'js-ray-keyboard-nav';

// Let the document know when the mouse is being used,
// so accessibility styling can be removed.
const addAccessibilityClass = () => {
      document.body.classList.add(ACCESSIBILITY_CLASS);
};

const removeAccessibilityClass = () => {
      document.body.classList.remove(ACCESSIBILITY_CLASS);
};

const setupAccessibility = () => {
  if (typeof document !== 'undefined') {
    document.addEventListener('mousedown', removeAccessibilityClass);
    document.addEventListener('keydown', addAccessibilityClass);
  }
}

export default setupAccessibility;
```

packages/core/test/global/accessibility.test.js
```diff
@@ -0,0 +1,51 @@
+import setupAccessibility from '../../src/global/js/accessibility';
+
+describe('accessibility', () => {
+  afterEach(() => {
+    document.body.classList = null;
+  });
+
+  test('it works with SSR', () => {
```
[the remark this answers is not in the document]
adamraider: server side rendering

packages/core/test/global/boot.test.js
```diff
@@ -7,6 +7,39 @@ const {
 } = require('../../src/global/js/boot');
 
 describe('boot', () => {
+  test('it works with SSR', () => {
```
designjockey:
```suggestion
  test('it works with server side rendering', () => {
```

packages/core/test/global/accessibility.test.js
designjockey:
```suggestion
  test('it works with server side rendering', () => {
```

packages/core/src/global/js/accessibility.js
```diff
@@ -0,0 +1,22 @@
...
+
+function setupAccessibility() {
+  if (typeof document !== 'undefined') {
```
kconst: Should this be `window.document`?

adamraider: It's not necessary.

adamraider: `typeof` will throw an error because something is undefined. It will just return `undefined`

adamraider: ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#Errors

packages/core/src/global/js/accessibility.js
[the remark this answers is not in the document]
adamraider: Yea we could use `const`s. If a function is going to be named I generally prefer to use the `function` keyword though as it's much easier to read imo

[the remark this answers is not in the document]
designjockey: Even so, you can get the functions out of the if block, to resolve the eslint alert.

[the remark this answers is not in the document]
adamraider: There was some reason I did it like this but I can't strictly remember atm 🤔

packages/core/src/global/js/accessibility.js
kconst: Awesome!  Totally forgot about this safety with typeof.

packages/core/src/global/js/accessibility.js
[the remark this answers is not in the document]
kconst: My guess was it lost scope somewhere that it was being used if you had to write it this way.  I like the route Ritesh suggested if it's doable.

docs/foundations/accessibility.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Color Ratio
+
+The WCAG state that to ensure readability, text and interactive elements should maintain a color contrast ratio of at least 4.5:1.
+
+[Web Content Accessibility Guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG)
+
+### Color Blindness
+
+The most common variety of colorblindness is red-green color blindness. It's important to **avoid** juxtaposing these color.
```
grmoon:
```suggestion
The most common variety of colorblindness is red-green color blindness. It's important to **avoid** juxtaposing these colors.
```

docs/foundations/accessibility.md
```diff
@@ -0,0 +1,33 @@
...
+
+Many people navigate the web using technologies such as screen readers, or are unable to use a traditional mouse. It's imperative that content and interactive elements are accessible by way of keyboard navigation, and clearly indicate where the user has focus.
+
+Do not remove focus states from elements unless you implement a suitable replacement.
+
+If the user is using a keyboard to navigate, Ray will apply a class to the document's body element that enables focus states on various components, otherwise these styles are hidden to avoid visual clutter.
+
+### Color Ratio
+
+The WCAG state that to ensure readability, text and interactive elements should maintain a color contrast ratio of at least 4.5:1.
```
grmoon:
```suggestion
The [Web Content Accessibility Guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG) state that to ensure readability, text and interactive elements should maintain a color contrast ratio of at least 4.5:1.
```

PR #164, opened 2019-07-09
fix(select): add space around arrow so doesn't overlap

cmugla:
I don't think there is a way to do this with pure css to show an ellipsis. 

BEFORE:
Screen Shot 2019-07-09 at 11 26 07 AM

AFTER:
Screen Shot 2019-07-09 at 11 25 51 AM


adamraider:
Nice, can you add a story so we can visually track?

PR #166, opened 2019-07-10
package: init core-react

adamraider:
Creates `@wework/ray-core-react` package, a React wrapper for all things ray-core.

Creates a jest project schema so we can more easily run the entire monorepo's test suite at once (and generate a coverage report!)

[Ray React Storybook](https://deploy-preview-166--ray-docs.netlify.com/core-react/storybook/?path=/story/*)


packages/core-react/src/components/Button/index.js
```diff
@@ -0,0 +1,41 @@
...
+import clsx from 'clsx';
+import PropTypes from 'prop-types';
+
+export default function Button({
+  Tag = 'button',
+  children,
+  type,
+  isCompact,
+  isDanger,
+  disabled,
```
i8ramin: should we standardize on `boolean` props? either `isSomething` or just `something`. For example, we have `disabled` and `isCompact` here. I personally prefer `compact`, `danger`, and `disabled`

adamraider: reason i left out disabled is because it's native api.. but yea you're approach makes sense. the prefixes might be overkill

seimithwework: I agree with ramin here :)

adamraider: cool, thanks for the input

packages/core-react/src/components/Button/index.js
```diff
@@ -0,0 +1,41 @@
...
+      })}
+      disabled={disabled}
+    >
+      {children}
+    </Tag>
+  );
+}
+
+Button.propTypes = {
+  type: PropTypes.oneOf(['primary', 'secondary', 'tertiary']),
```
seimithwework: We can talk some more about Typescript vs Proptypes, but for the time being I think this is fine. I've incorporated Typescript because my team and the projects consuming our components also use TS (however, you can totally use it even if your app doesn't use TS). 

I've found using TS helpful because it's actually helped with debugging purposes while creating functional components. 

You might like this article: https://medium.com/@jtomaszewski/why-typescript-is-the-best-way-to-write-front-end-in-2019-feb855f9b164

I found some articles that go over the pros and cons of using TypeScript as I would't do great justice in explaining

packages/core-react/src/components/Button/index.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import clsx from 'clsx';
```
cmugla: cool

packages/core-react/src/components/Card/index.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import clsx from 'clsx';
+import PropTypes from 'prop-types';
+
+export default function Card({ heading, body, img, row }) {
+  return (
+    <div className={clsx('ray-card', { 'ray-card--row': row })}>
```
cmugla: will need to add link support
```suggestion
    <div className={clsx('ray-card', { 'ray-card--row': row, 'ray-card--link': link })}>
```

packages/core-react/README.md
```diff
@@ -0,0 +1,13 @@
+# [Ray React Components](https://ray.wework.com/)
+
+Resources for building interfaces with WeWork's design system.
```
cmugla: since this is open source, should probably add a disclaimer that this is a work in progress and probably not production ready yet

adamraider: Should be denoted by the "alpha" release tag

packages/core-react/src/components/Card/index.js
adamraider: Good point

adamraider: I think I might do that in a follow-up since this pr is mainly about getting the infra off the ground

packages/core-react/src/components/Button/index.js
adamraider: Don't think typescript is really necessary here, I'm not a fan of the bloat. Thank you for the suggestion though.

PR #169, opened 2019-07-10
fix(checkbox): squishing on small containers

cmugla:
BEFORE:
Screen Shot 2019-07-10 at 5 06 07 PM

AFTER:
Screen Shot 2019-07-10 at 7 04 45 PM


packages/core/src/components/radio-checkbox/_radio-checkbox.scss
```diff
@@ -25,7 +25,7 @@
         content: '';
         background-color: $ray-color-white;
         border: $ray-radio-checkbox-border-width solid $ray-color-black;
-        width: $ray-radio-checkbox-size;
+        min-width: $ray-radio-checkbox-size;
```
adamraider: can we also set width here? to really make sure that its always a square

cmugla: fo sho

PR #179, opened 2019-07-18
feat(core-react): add text field 

adamraider:


packages/core-react/src/components/TextField/index.js
```diff
@@ -0,0 +1,59 @@
...
+  disabled,
+  required,
+  value,
+  error,
+  compact,
+  label,
+  ...rest
+}) {
+  const wrapperClass = clsx('ray-text-field', {
+    'ray-text-field--has-value': value,
```
i8ramin:
```suggestion
    'ray-text-field--has-value': !!value,
```

packages/core-react/src/components/TextField/index.js
```diff
@@ -0,0 +1,59 @@
...
+  placeholder,
+  disabled,
+  required,
+  value,
+  error,
+  compact,
+  label,
+  ...rest
+}) {
+  const wrapperClass = clsx('ray-text-field', {
```
i8ramin: Should this also handle textarea?

    'ray-text-field': !textarea,
    'ray-text-area': textarea,

packages/core-react/src/components/TextField/index.js
```diff
@@ -0,0 +1,59 @@
...
+        placeholder={placeholder}
+        disabled={disabled}
+        {...rest}
+      />
+      {label && (
+        <label className="ray-text-field__label" htmlFor={id}>
+          {label}
+        </label>
+      )}
+    </div>
```
cmugla: should we also include the error message?

adamraider: i was considering it... this pr is still wip

packages/core-react/src/components/TextField/index.js
adamraider: maaybeee.

packages/core-react/src/components/TextField/index.js
adamraider: why ?

PR #181, opened 2019-07-23
feat(chip): add / modify chip component

gmharrison:
A continuation of https://github.com/wework/ray/pull/10 that adds icon support for chips

https://cl.ly/554d75534195


docs/html/chip/chip--with-icon.html
```diff
@@ -0,0 +1,38 @@
...
...C5.64661382,1.27983585 3.70319409,3.118366 3.57275624,5.46812282 L3.47202024,7.28041082 Z M4.10830824,14.0800908 L3.53749224,14.0800908 L3.32418024,14.7200908 L3.89493224,14.7200908 L4.10830824,14.0800908 Z M12.0357962,14.7200908 L11.8224842,14.0800908 L11.2517322,14.0800908 L11.4650442,14.7200908 L12.0357962,14.7200908 Z M14.7199562,8.96009082 C14.7199562,8.25316634 14.1468807,7.68009082 13.4399562,7.68009082 C12.7330318,7.68009082 12.1599562,8.25316634 12.1599562,8.96009082 L12.1599562,10.2400908 C12.1591451,10.946679 11.5865444,11.5192796 10.8799562,11.5200908 L5.11995624,11.5200908 C4.41336803,11.5192796 3.84076741,10.946679 3.83995624,10.2400908 L3.83995624,8.96009082 C3.83995624,8.25316634 3.26688072,7.68009082 2.55995624,7.68009082 C1.85303176,7.68009082 1.27995624,8.25316634 1.27995624,8.96009082 C1.28035109,9.32233784 1.43636572,9.66695463 1.70830824,9.90626682 L1.91995624,10.0969868 L1.91995624,10.8800908 C1.9215434,12.2932818 3.06676522,13.4385037 4.47995624,13.4400908 L11.5199562,13.4400908 C12.9331472,13.4385037 14.0783691,12.2932818 14.0799562,10.8800908 L14.0799562,10.0969868 L14.2916042,9.90626682 C14.5635468,9.66695463 14.7195614,9.32233784 14.7199562,8.96009082 Z"
```
adamraider: thats an svg alright

packages/core/src/components/chip/_chip.scss
```diff
@@ -0,0 +1,65 @@
...
+    color: $ray-color-white;
+  }
+
+  .#{$ray-class-prefix}chip__icon {
+    vertical-align: middle;
+    display: none;
+  }
+
+  .#{$ray-class-prefix}chip__icon--active {
+    display: inline-block;
```
adamraider: what's this used for?

packages/core/src/components/chip/_chip.scss
```diff
@@ -0,0 +1,65 @@
...
+  display: none;
+  vertical-align: middle;
+}
+
+.#{$ray-class-prefix}chip--micro {
+  padding: 0 0.75rem;
+}
+
+.#{$ray-class-prefix}chip__text {
+  margin-left: 10px;
```
adamraider: prefer to use rem here please

packages/core/stories/chip.stories.js
```diff
@@ -0,0 +1,156 @@
...
...95 2.56056439,0.430040593 2.56,0.96 L2.56,15.36 L1.6,15.36 C1.42326888,15.36 1.28,15.5032689 1.28,15.68 C1.28,15.8567311 1.42326888,16 1.6,16 L14.4,16 C14.5767311,16 14.72,15.8567311 14.72,15.68 C14.72,15.5032689 14.5767311,15.36 14.4,15.36 Z M9.6,15.36 L6.4,15.36 L6.4,10.88 L9.6,10.88 L9.6,15.36 Z M12.8,15.36 L10.24,15.36 L10.24,10.88 C10.24,10.5265378 9.95346224,10.24 9.6,10.24 L6.4,10.24 C6.04653776,10.24 5.76,10.5265378 5.76,10.88 L5.76,15.36 L3.2,15.36 L3.2,0.96 C3.20035211,0.783414936 3.34341494,0.640352115 3.52,0.64 L12.48,0.64 C12.6565851,0.640352115 12.7996479,0.783414936 12.8,0.96 L12.8,15.36 Z M10.88,5.76 L5.12,5.76 C4.76653776,5.76 4.48,6.04653776 4.48,6.4 L4.48,7.68 C4.48,8.03346224 4.76653776,8.32 5.12,8.32 L10.88,8.32 C11.2334622,8.32 11.52,8.03346224 11.52,7.68 L11.52,6.4 C11.52,6.04653776 11.2334622,5.76 10.88,5.76 Z M10.88,7.68 L5.12,7.68 L5.12,6.4 L10.88,6.4 L10.88,7.68 Z M10.88,1.92 L5.12,1.92 C4.76653776,1.92 4.48,2.20653776 4.48,2.56 L4.48,3.84 C4.48,4.19346224 4.76653776,4.48 5.12,4.48 L10.88,4.48 C11.2334622,4.48 11.52,4.19346224 11.52,3.84 L11.52,2.56 C11.52,2.20653776 11.2334622,1.92 10.88,1.92 Z M10.88,3.84 L5.12,3.84 L5.12,2.56 L10.88,2.56 L10.88,3.84 Z"
```
adamraider: is there a less crazy svg we could use so the test file isnt so cluttered lol

adamraider:
Should probably add keyboard navigation (make sure focus states work and ability to toggle with space bar)

packages/core/src/components/chip/_chip.scss
adamraider: this will not work if we want to put the icon on the right side of the text :/ can we figure a location agnostic approach? This will be especially important for RTL languages

packages/core/src/components/chip/_chip.scss
```diff
@@ -0,0 +1,65 @@
...
+  border: $ray-chip-border-width solid $ray-color-blue-50;
+  background-color: $ray-color-white;
+  padding: $ray-spacing-xs $ray-spacing-md;
+  border-radius: $ray-chip-height;
+  color: $ray-color-blue-50;
+  display: inline-block;
+  cursor: pointer;
+
+  &:not(:last-child) {
+    margin-right: $ray-spacing-xs;
```
adamraider: Please verify this is compatible with RTL (see other components for reference like buttons)

packages/core/test/components/chip.test.js
```diff
@@ -0,0 +1,62 @@
+import html from 'nanohtml';
+import Chip from '../../src/components/chip';
+
+function triggerEvent(el, type) {
```
adamraider: cant remember if this is used elsewhere but we maaaay want to abstract into a util, definitely not blocking on this

packages/core/src/components/chip/_chip.scss
gmharrison: ah okay, yeah i was curious about your thoughts on this. i'm fine with just assuming that whoever is using the component can decide how they want their icons to behave.

packages/core/stories/chip.stories.js
gmharrison: yeah i can use a circle or something

packages/core/src/components/chip/_chip.scss
gmharrison: removed icon related styles/classes

docs/html/chip/chip--with-icon.html
gmharrison: lol i changed it to be smaller

packages/core/src/components/chip/_chip.scss
gmharrison: styling added to support RTL languages

packages/core/stories/chip.stories.js
gmharrison: should be a small square now

packages/core/src/components/chip/_chip.scss
gmharrison: fixed to be compatible with RTL

gmharrison:
@adamraider added keyboard nav support! thanks for catching that

packages/core/src/components/chip/index.js
```diff
@@ -0,0 +1,78 @@
...
+      this._root.classList.remove(this.constructor.cssClasses.ACTIVE);
+    }
+  }
+
+  onClick = () => {
+    this.state.active = !this.state.active;
+    this.assignClasses();
+  };
+
+  onSpace = e => {
```
adamraider: the linter should've caught this 🤔 (no variable names < 2 chars)

packages/core/src/components/chip/index.js
```diff
@@ -0,0 +1,78 @@
...
+    if (e.code === 'Space') {
```
adamraider: image

pls add test coverage for this method

adamraider: .code lacks IE support https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code ;_;

gmharrison: added a check for the actual `keyCode`

gmharrison: added tests

packages/core/src/components/chip/index.js
gmharrison: made the var name `event`

packages/core/src/components/chip/index.js
gmharrison: added a check for the actual `keyCode`

gmharrison: lol jk that's deprecated.

_now_ i added a check for `key` which is supported by IE and also not deprecated

packages/core/src/components/chip/index.js
```diff
@@ -0,0 +1,78 @@
...
+    this.state = {
+      active: this._root.classList.contains(this.constructor.cssClasses.ACTIVE)
+    };
+
+    this._bindEventListeners();
+    this.assignClasses();
+  }
+
+  _bindEventListeners() {
+    this._root.addEventListener('mousedown', this.onClick);
```
cmugla: this came up for another component I can't quite remember, but should we add this for `touchstart` as well? Did you test on your phone?

gmharrison: great catch! it looks like both the `touch` events and `click` events retain hover styling when clicking on mobile. so i added the hover style to non mobile screens only. 

here's a lil demo https://cl.ly/16bf6c68733d

packages/core/src/components/chip/_chip.scss
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @include ray-breakpoint(desktop) {
+    &:hover {
+      background-color: $ray-color-blue-10;
+      color: $ray-color-blue-50;
+    }
+  }
+
+  &:focus {
```
adamraider: Please put this behind the accessibility class (see https://github.com/wework/ray/blob/master/packages/core/src/components/button/_button.scss#L77 for an example).

We do this so the focus states dont appear on tap or click, which can be confusing for some people as they wont go away until the user focuses on another element

packages/core/stories/chip.stories.js
```diff
@@ -0,0 +1,84 @@
...
+              fill="#FFFFFF"
+            />
+          </svg>
+          <span className="ray-chip__text">Offices</span>
+        </div>
+      </div>
+    </div>
+  );
+});
+storiesOf('Chip', module).add('micro', () => {
```
adamraider: Please add a story for RTL

gmharrison: a story or an example in the docs? the cards for example have an RTL example in the docs, but do not have an associated RTL story

gmharrison: added

packages/core/stories/chip.stories.js
```diff
@@ -21,6 +21,19 @@ storiesOf('Chip', module).add('default', () => {
...
+      <div tabIndex="1" className="ray-chip">
+        Neighborhoods
+      </div>
+      <div tabIndex="1" className="ray-chip ray-chip--active">
+        Offices
+      </div>
+    </div>
+  );
+});
 storiesOf('Chip', module).add('micro', () => {
```
adamraider: nit: you can chain these .add statements

packages/core/src/components/chip/_chip.scss
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @include ray-breakpoint(desktop) {
+    &:hover {
+      background-color: $ray-color-blue-10;
+      color: $ray-color-blue-50;
+    }
+  }
+
+  .#{$ray-accessibility-class} &:focus {
```
adamraider: need to add

```css
outline: 0;
```

to remove this effect:

image

docs/html/chip/chip.html
```diff
@@ -0,0 +1,3 @@
+<div class="ray-chip">
```
adamraider: can you please add the tab index attribute to the html examples?

gmharrison: yup

PR #183, opened 2019-07-25
fix: IE11 issue

cmugla:
babel doesn't transpile `new Event()` for IE11 :(

so let's do it ourselves using [`createEvent`](https://developer.mozilla.org/en-US/docs/Web/API/Document/createEvent) - tested in Browserstack IE11


adamraider:
is there anyway to add a plugin or something to babel for this?

cmugla:
>is there anyway to add a plugin or something to babel for this?

I couldn't find one

PR #185, opened 2019-07-26
fix: adds set method and change event for chip

gmharrison:
Emits a change event in `onClick` and `onSpace` so consumers can set event listeners for just the `change` event. Also adds a `set` method that allows consumers to set the active state of the chip

### DIFF
--- a/packages/core/src/components/chip/index.js
+++ b/packages/core/src/components/chip/index.js
@@ -57,12 +57,14 @@ class Chip {
   onClick = () => {
     this.state.active = !this.state.active;
     this.assignClasses();
+    this.emitChange();
   };
 
   onSpace = event => {
     if (event.code === 'Space' || event.key === ' ') {
       this.state.active = !this.state.active;
       this.assignClasses();
+      this.emitChange();
     }
   };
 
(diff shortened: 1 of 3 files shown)


packages/core/src/components/chip/index.js
```diff
@@ -73,6 +75,17 @@ class Chip {
...
     this.constructor.instances.delete(this._root);
   }
+
+  emitChange() {
+    const newEvent = document.createEvent('HTMLEvents');
+    newEvent.initEvent('change', true, true);
+    this._root.dispatchEvent(newEvent);
+  }
+
+  set(isActive) {
```
cmugla: should we also `emitChange` here too?

gmharrison: hmmm i think that makes sense

gmharrison: updated!

gmharrison:
@adamraider tests added

gmharrison:
@adamraider updated

adamraider:
Don't have docs around the event but its ok for now

PR #191, opened 2019-08-05
feat(core-react): work on react card and aspect ratios

i8ramin:


packages/core-react/src/components/Card/index.js
```diff
@@ -2,10 +2,34 @@ import React from 'react';
 import clsx from 'clsx';
 import PropTypes from 'prop-types';
 
-export default function Card({ heading, body, img, row }) {
+import rayExports from '@wework/ray-core/scss/ray-exports.scss';
+
+const aspectRatios = (rayExports['ray-image-aspect-ratios'] || '').split(' ');
```
adamraider:
```suggestion
const ASPECT_RATIOS = (rayExports['ray-image-aspect-ratios'] || '').split(' ');
```

packages/core-react/src/components/Card/index.js
```diff
@@ -18,5 +42,12 @@ Card.propTypes = {
   heading: PropTypes.node,
   body: PropTypes.node,
   img: PropTypes.node,
-  row: PropTypes.bool
+  row: PropTypes.bool,
+  aspectRatio: PropTypes.oneOf(aspectRatios),
```
adamraider:
```suggestion
  aspectRatio: PropTypes.oneOf(ASPECT_RATIOS),
```

packages/core/src/ray-exports.scss
```diff
@@ -1,5 +1,10 @@
 @import 'global/variables';
 
+$aspect-ratios: ();
```
adamraider: why is this part needed?

packages/core-react/src/components/Button/index.js
```diff
@@ -9,14 +9,16 @@ export default function Button({
...
   ...props
 }) {
   return (
     <Tag
       {...props}
       className={clsx('ray-button', `ray-button--${type}`, {
         'ray-button--compact': compact,
-        'ray-button--danger': danger
+        'ray-button--danger': danger,
+        [`${className}`]: !!className
```
adamraider: - do you need to interpolate?
- do you need to type coerce? 

```suggestion
        [className]: className
```

packages/core/src/ray-exports.scss
i8ramin: Took from doc examples. I believe you need to initialize it to something so can `append` to it

packages/core-react/src/components/Button/index.js
i8ramin: hmm, good point

packages/core-react/src/components/Card/index.js
```diff
@@ -2,10 +2,34 @@ import React from 'react';
...
+  ...props
+}) {
   return (
-    <div className={clsx('ray-card', { 'ray-card--row': row })}>
-      <div className="ray-card__image ray-image ray-image--16by9">{img}</div>
+    <div
+      {...props}
+      className={clsx('ray-card', {
+        'ray-card--row': row,
+        [`${className}`]: !!className
```
i8ramin:
```suggestion
        [className]: className
```

packages/core-react/config/rollup.config.js
```diff
@@ -17,6 +19,18 @@ module.exports = {
     }
   },
   plugins: [
+    scss({
```
adamraider: I would strongly prefer not to add css processing to our build config. I find that things just get extremely confusing when your js bundler starts processing styles.

packages/core-react/src/components/Card/index.js
```diff
@@ -2,10 +2,34 @@ import React from 'react';
 import clsx from 'clsx';
 import PropTypes from 'prop-types';
 
-export default function Card({ heading, body, img, row }) {
+import rayExports from '@wework/ray-core/scss/ray-exports.scss';
```
adamraider: Can we instead add this to the javascript settings and import from there?


https://github.com/wework/ray/blob/master/packages/core/src/global/js/settings.js

packages/core-react/src/components/Card/index.js
```diff
@@ -2,10 +2,34 @@ import React from 'react';
...
+    <div
+      {...props}
+      className={clsx('ray-card', {
+        'ray-card--row': row,
+        [className]: className
+      })}
+    >
+      <div
+        className={clsx('ray-card__image', 'ray-image', {
+          [`ray-image--${aspectRatio}`]: !!aspectRatio
```
adamraider: Is the boolean coercion necessary? 
```suggestion
          [`ray-image--${aspectRatio}`]: aspectRatio
```

packages/core-react/src/components/Card/index.js
```diff
@@ -2,10 +2,34 @@ import React from 'react';
...
+  ...props
+}) {
   return (
-    <div className={clsx('ray-card', { 'ray-card--row': row })}>
-      <div className="ray-card__image ray-image ray-image--16by9">{img}</div>
+    <div
+      {...props}
+      className={clsx('ray-card', {
+        'ray-card--row': row,
+        [className]: className
```
adamraider: btw you dont have to interpolate here, you can just add className as an additional argument like so:

```js
clsx('ray-card', { 'ray-card--row': row }, className)
```

packages/core-react/src/components/Card/index.js
i8ramin: ahh, didnt know about that. will look into it. Although, dont see any advantage in sharing info between css/jss this way? to reduce duplication.

packages/core-react/config/rollup.config.js
i8ramin: this is mainly to get the scss exports to work. we should discuss that, since main motivation was to avoid duplicating css config in scss and JS

cmugla (approved):

PR #201, opened 2019-08-15
Create Bug_report.md

wuweiweiwu:


adamraider:
not a fan of these as they add unnecessary clutter to the github (imo)

ty for the suggestion

wuweiweiwu:
sounds good!

PR #202, opened 2019-08-15
Create Feature_request.md

wuweiweiwu:


adamraider:
i dont really think we need these, they just add clutter to our github imo

adamraider:
i appreciate the suggestion though

wuweiweiwu:
makes sense!

PR #204, opened 2019-08-22
refactor: corrects spelling of pikachu

taekyunggg:
corrects spelling from 'pikatchu' to 'pikachu'


omervk:
FYI it should be `פיקצ'ו` and not `פיקציו`. cc @iainmcgin ;P

PR #210, opened 2019-08-28
fix(core-react): Exporting existing components for use!

wuweiweiwu:
🎂


wuweiweiwu:
I can own adding some tests!

seimithwework:
@adamraider I just added some small tests and wanted to make sure you're ok with exporting now! :-)

PR #211, opened 2019-08-29
Add Tabs to React-Product

jgbenito7:
JIRA Ticket: [WEB-517](https://jira.weworkers.io/browse/WEB-517)

Arch Review Doc: https://docs.google.com/document/d/1K4UBh2eJN_Ey4EV4siBEmMxcoJqAMa6bm2NlCliw8N4/edit#

**Change Details**

- Adds a Ray-Product Tabs component that allows for easy creation of tabs in a web page.

This PR also contains A LOT of infrastructure additions to make sure that things will compile correctly. 

**Screenshots, gifs, references**

**Packages / Dependencies**

None

**How to test this feature**

- `yarn storybook`
- View the Tabs component and its stories
- Try out the a11y stuff!
   - focus a tab
   - press ArrowLeft, ArrowRight, Home and End to see what happens

---
I Reviewed the Requirements:
- [x] PR is labeled correctly (`ready for review`, `package: ...`, `app: ...`, etc)
- [x] CI passes


adamraider:
Please run this by @danaballasy

packages/product-react/package.json
```diff
@@ -16,7 +17,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@babel/preset-typescript": "^7.3.3",
-    "@wework/ray-core": "^1.9.0",
+    "@wework/ray-core": "^1.8.5",
```
adamraider: not sure what happened here

packages/product-react/src/Button/__snapshots__/Button.test.tsx.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
adamraider: Can we remove these please

packages/product-react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,220 @@
+import classnames from 'classnames';
+import * as React from 'react';
+// import { Button } from '../';
+import { Button } from '@wework/ray-core-react';
+import { TabContext } from './context';
+
+// console.log({ stuff });
```
adamraider: 🔪 

pls remove commented out code

packages/product-react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const { tabs, firstFocusedId, buttonsFocused } = props;
+  const [focusedTab, setFocusedTab] = React.useState(firstFocusedId);
+  const [isMounted, setIsMounted] = React.useState(false);
+
+  const tabRefs: ITabRefs = tabs.reduce((acc, { id }) => {
+    return { ...acc, [id]: React.useRef<HTMLButtonElement>(null) };
+  }, {});
+
+  React.useEffect(() => {
+    if (!isMounted) {
```
adamraider: Is this used anywhere?

jgbenito7: Yeah, this is to make the hook listen to changes to `focusedTab` but ignore the first render (mount).

packages/product-react/src/Tabs/TabView.scss
```diff
@@ -0,0 +1,3 @@
+.ray-product__tab-view:focus {
+  outline: none;
```
seimithwework: I want to call this out to @danaballasy and @simon that we're removing the blue outline of the container. I know that there are very strong opinions on keeping it due to accessibility, but since this would be for the container below the tab, I suggested that we remove it.

packages/product-react/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@wework/ray-core-react';
```
seimithwework: Questionable! Not sure if we need this, but if it borks without it, let's figure it out.

packages/product-react/src/Tabs/Tabs.scss
```diff
@@ -0,0 +1,66 @@
+$easing: cubic-bezier(0.694, 0.0482, 0.335, 1);
+$ray-color-black: black;
+$ray-color-blue-50: blue;
```
seimithwework: Are we not able to use `Ray-core` variables here instead 😛?

adamraider: yes pls!

packages/product-react/src/Tabs/Tabs.scss
```diff
@@ -0,0 +1,66 @@
...
+$hover-animation-time: 0.33s;
+
+@mixin tab-border($border-size, $color) {
+  content: '';
+  position: absolute;
+  top: 100%;
+  height: $border-size;
+  left: 0;
+  width: 100%;
+  background-color: $color;
```
seimithwework: Nit: to alphabetize

packages/product-react/src/Tabs/Tabs.scss
```diff
@@ -0,0 +1,66 @@
...
+
+  & button {
+    width: 100%;
+  }
+}
+
+.ray-tabs__button-wrapper--active {
+  &::after {
+    @include tab-border($hover-border-size, $ray-color-blue-50);
+    transition: unset; // stylelint-disable-line property-blacklist
```
seimithwework: Nit: to remove comment

packages/product-react/src/Tabs/Tabs.scss
```diff
@@ -0,0 +1,66 @@
...
+  width: 100%;
+}
+
+.ray-tabs__button-wrapper {
+  position: relative;
+  flex: 1;
+
+  @include underline-hover($hover-border-size, $hover-animation-time);
+
+  & button {
```
seimithwework: Can we add a class instead of targeting the tag?

packages/product-react/src/Tabs/Tabs.scss
```diff
@@ -0,0 +1,66 @@
+$easing: cubic-bezier(0.694, 0.0482, 0.335, 1);
+$ray-color-black: black;
+$ray-color-blue-50: blue;
+$hover-border-size: 2px;
+$hover-animation-time: 0.33s;
```
seimithwework: @adamraider Are you hard-pressed on seconds? What do you think about milliseconds?

packages/product-react/src/Tabs/TabView.scss
jgbenito7: Yeah we need an answer on this one

packages/product-react/src/Tabs/Tabs.scss
jgbenito7: If we want to enforce this, should we add stylelint rules?

packages/product-react/src/Tabs/Tabs.scss
jgbenito7: transition is actually blacklisted in this repo, yet the stylelint disable comment is used in several places. Should this blacklist be removed then @adamraider ?

adamraider: I'd like to maintain the blacklist until we work with design to figure out an animation/transition strategy. The disable comments you're seeing are actually in the `docs-app` package. In other words these dont actually affect the Ray library.

packages/product-react/src/Tabs/Tabs.scss
adamraider: I don't feel strongly either way

packages/product-react/src/Tabs/Tabs.scss
adamraider: I'd prefer this as well

packages/product-react/src/Tabs/Tabs.scss
adamraider: I don't feel strongly about it, make a gh issue if you'd like

packages/product-react/index.d.ts
adamraider: What does this do 🤔

wuweiweiwu:
is this accessible? can I tab navigate the component and press <enter> to change tabs?

https://www.w3.org/TR/wai-aria-practices/#tabpanel

a reference https://reakit.io/docs/tab/ as reakit was something we were going to base accessible components off of

packages/product-react/index.d.ts
wuweiweiwu: i believe it just says that the `@wework/ray-core-react` module has any type

wuweiweiwu: but this should be removed as the most current ray core react has typings

packages/product-react/src/Tabs/TabView.scss
wuweiweiwu: my opinion is that we should not be removing it. there is still a need for users to be able to visibly detect the body of the tab is focused

packages/product-react/src/Tabs/TabView.test.tsx
```diff
@@ -0,0 +1,77 @@
+import Enzyme, { configure } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
```
wuweiweiwu: this should be in a global `setupTests.js` file similar to https://github.com/WeConnect/wework-web-monorepo/blob/master/jest/setupTests.ts

packages/product-react/src/Tabs/TabView.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+import React from 'react';
+import { TabContext } from './context';
+import { TabView } from './TabView';
+
+configure({ adapter: new Adapter() });
+
+interface IContextProviderProps {
```
wuweiweiwu: nit: prefer `type` over `interface`

wuweiweiwu: https://medium.com/@martin_hotell/10-typescript-pro-tips-patterns-with-or-without-react-5799488d6680 #19 has some really good points

packages/product-react/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import '@wework/ray-core/scss/ray-core.scss';
```
wuweiweiwu: this should be removed. 

the responsibility of loading the styles is on the consumer

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
```
wuweiweiwu: this should be scoped to product react. as other packages are not on typescript yet

package.json
```diff
@@ -37,7 +42,8 @@
     "lerna": "3.13.4",
     "lint-staged": "8.1.4",
     "prettier": "1.16.4",
-    "stylelint": "^9.10.1"
+    "stylelint": "^9.10.1",
+    "typescript": "^3.6.2"
```
wuweiweiwu: i think this should be scoped to `product-react`

packages/product-react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+      if (currentIndex === 0) {
+        return getLastTabId(tabs);
+      }
+
+      return getIdAt(tabs, currentIndex - 1);
+    });
+  };
+
+  const focusRoving = ({ key }: { key: string }) => {
```
wuweiweiwu: this should be typed as `React.KeyboardEvent`

packages/product-react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,217 @@
...
+    switch (key) {
```
wuweiweiwu: should there be a `default` case to unfocus?

if the user pressed `tab` it should focus something else on the page

packages/product-react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,217 @@
...
+    if (!isMounted) {
+      return setIsMounted(true);
+    }
+
+    const tabRef = tabRefs[focusedTab];
+
+    if (tabRef.current) {
+      tabRef.current.focus();
+    }
+  }, [focusedTab]);
```
wuweiweiwu: i think as a rule of thumb with effect hooks any values you reference within should be a dependency or else you'll end up with stale values

packages/product-react/index.d.ts
jgbenito7: Does it? I'll try bumping the version

packages/product-react/src/Tabs/TabView.test.tsx
jgbenito7: If we want to enforce this we should set up linting for it. Interface has been the paradigm i've seen us use so far

packages/product-react/src/Tabs/Tabs.tsx
jgbenito7: Can you be more specific? are you talking about isMounted?

packages/product-react/src/Tabs/Tabs.tsx
jgbenito7: No need to handle that in here, that should happen just by browser default.

packages/product-react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+export interface IProps {
+  tabs: ITab[];
+  defaultActiveTab?: string;
+  onSelect?: (selectedTabId: string) => void;
+  compact?: boolean;
+}
+
+export const Tabs: React.FC<IProps> = props => {
+  const { tabs, onSelect, compact } = props;
```
seimithwework: Are we missing `defaultActiveTab` here too?

packages/product-react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { TabContext } from './context';
+
+import './Tabs.scss';
+
+interface ITabRefs {
+  [key: string]: React.RefObject<HTMLButtonElement>;
+}
+
+export interface ITab {
+  id: string;
```
seimithwework: I'm assuming that the ID will most likely be a string as well and somebody won't just use their index. 😂

package.json
```diff
@@ -21,6 +21,11 @@
     "@babel/core": "7.4.4",
     "@babel/plugin-proposal-class-properties": "7.4.4",
     "@babel/preset-env": "7.3.4",
+    "@types/classnames": "^2.2.9",
```
seimithwework: We should align with the Ray team with using `clsx` now.

packages/product-react/src/components/Tabs/TabView.scss
```diff
@@ -0,0 +1,3 @@
+.ray-product__tab-view:focus {
+  outline: none;
```
seimithwework: I know people are going to ask about this and I know that this (http://www.outlinenone.com/) exists... But I stand by NOT having the blue accessibility outline around the TabView.

jgbenito7: I kind of like the outline for tabbing, but not for mouse clicks.

seimithwework: Oh yeah I'm good on the tabs, but now that tabview itself :-P

packages/product-react/src/components/Tabs/Tabs.scss
```diff
@@ -0,0 +1,66 @@
...
+) {
+  position: relative;
+  text-decoration: none;
+
+  &::after {
+    @include tab-border($border-size, $ray-color-black);
+
+    transform-origin: center;
+    transform: scaleX(0);
+    transition: transform $transition-timing $easing; // stylelint-disable-line property-blacklist
```
jgbenito7: transition here seems much more natural for a hover in/out animation than using keyframes. I have't even found a working solution yet using keyframes to do this... What do we think about this?

PR #213, opened 2019-08-29
feat(core-react): add typings and Tag prop to Container, Grid & Cell

Morgantheplant:
changes
- add Tag prop for allowing for more semantic html tags in layout components
- adds typings for Container, Grid, & Cell

### DIFF
--- a/packages/core-react/index.d.ts
+++ b/packages/core-react/index.d.ts
@@ -8,4 +8,24 @@ export type TextFieldProps = {
 
 declare const TextField: React.FC<TextFieldProps>;
 
-export { TextField };
+export type ContainerProps = {
+  Tag?: React.ElementType<any>;
+} & React.HTMLProps<HTMLElement>;
+
+declare const Container: React.FC<ContainerProps>;
+
+export type GridProps = {
+  Tag?: React.ElementType<any>;
+} & React.HTMLProps<HTMLElement>;
+
+declare const Grid: React.FC<GridProps>;
+
+export type CellProps = {
+  push?: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+  span?: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+  Tag?: React.ElementType<any>;
+} & React.HTMLProps<HTMLElement>;;
+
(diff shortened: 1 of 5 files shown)


packages/core-react/index.d.ts
```diff
@@ -9,3 +9,29 @@ export type TextFieldProps = {
 declare const TextField: React.FC<TextFieldProps>;
 
 export { TextField };
+
+export type ContainerProps = {
+  Tag?: React.ElementType<any>;
```
Morgantheplant: following type from here: https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/ADVANCED.md#as-props-passing-a-component-to-be-rendered

Note: I am following the convention in this repo of naming the prop `Tag`. I do like the `as` prop naming convention for passed in tags. I can update if that is preferred.

packages/core-react/index.d.ts
```diff
@@ -9,3 +9,29 @@ export type TextFieldProps = {
 declare const TextField: React.FC<TextFieldProps>;
 
 export { TextField };
```
wuweiweiwu: can we put all our exports in one place? this is going to be hard to track as the # of components grow

packages/core-react/index.d.ts
wuweiweiwu: @adamraider any strong opinions about `as` vs. `Tag` prop?

packages/core-react/index.d.ts
```diff
@@ -9,3 +9,29 @@ export type TextFieldProps = {
...
+
+export type CellProps = {
+  push?: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+  span?: 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+  Tag?: React.ElementType<any>;
+} & React.HTMLProps<HTMLElement>;;
+
+declare const Cell: React.FC<CellProps>;
+
+export { Cell };
```
wuweiweiwu: cell shouldnt be exported anymore as it is a static field on Grid

Morgantheplant:
instead of adding functionality here I am going to close this and instead move over the existing component into ray-core. This will be split into 2 separate PRs.
see: https://jira.weworkers.io/browse/WEB-754

PR #214, opened 2019-08-30
chore(core-react): add typings for button

jgbenito7:
Add typings for Button in core-react.

### DIFF
--- a/packages/core-react/index.d.ts
+++ b/packages/core-react/index.d.ts
@@ -8,4 +8,16 @@ export type TextFieldProps = {
 
 declare const TextField: React.FC<TextFieldProps>;
 
-export { TextField };
+export type ButtonProps = {
+  Tag?: 'button' | 'a';
+  children?: React.ReactNode;
+  type?: 'primary' | 'secondary' | 'tertiary';
+  compact?: boolean;
+  danger?: boolean;
+  disabled?: boolean;
+  className?: string;
+};
+
+declare const Button: React.FC<ButtonProps>;
+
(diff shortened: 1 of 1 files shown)


packages/core-react/index.d.ts
```diff
@@ -8,4 +8,16 @@ export type TextFieldProps = {
 
 declare const TextField: React.FC<TextFieldProps>;
 
-export { TextField };
+export type ButtonProps = {
```
jgbenito7: Should we start a `types` folder for the definitions to keep things more organized?

adamraider: yea

packages/core-react/index.d.ts
```diff
@@ -8,4 +8,16 @@ export type TextFieldProps = {
...
+  Tag?: 'button' | 'a';
+  children?: React.ReactNode;
+  type?: 'primary' | 'secondary' | 'tertiary';
```
seimithwework: I'm ok with this for now, but I think we should definitely update these to enums in the future.

packages/core-react/index.d.ts
```diff
@@ -8,4 +8,16 @@ export type TextFieldProps = {
...
+  compact?: boolean;
+  danger?: boolean;
+  disabled?: boolean;
+  className?: string;
+};
```
seimithwework: I think we can add  `& React.HTMLProps<HTMLButtonElement>` at the end here too. (I'll add and push!)

PR #216, opened 2019-09-03
chore(core-react): add typings for radio

seimithwework:
Add types for Radios

### DIFF
--- a/packages/core-react/index.d.ts
+++ b/packages/core-react/index.d.ts
@@ -9,10 +9,10 @@ export type TextFieldProps = {
 declare const TextField: React.FC<TextFieldProps>;
 
 export type RadioProps = {
-  className?: string,
-  id: string,
-  label?: React.ReactNode
-  name?: string,
+  className?: string;
+  id?: string;
+  label?: React.ReactNode;
+  name?: string;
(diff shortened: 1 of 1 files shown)


packages/core-react/index.d.ts
```diff
@@ -8,6 +8,15 @@ export type TextFieldProps = {
 
 declare const TextField: React.FC<TextFieldProps>;
 
+export type RadioProps = {
+  className?: string,
+  id: string,
```
jamesdamild: I think maybe id should be optional

seimithwework: Ah ok, I was on the edge about that too

packages/core-react/index.d.ts
```diff
@@ -8,4 +8,13 @@ export type TextFieldProps = {
 
 declare const TextField: React.FC<TextFieldProps>;
 
-export { TextField };
+export type RadioProps = {
+  className?: string,
+  id: string,
+  label?: React.ReactNode
```
seimithwework: My bad - forgot a comma!

packages/core-react/index.d.ts
jamesdamild: I think each line should be finished by ; not ,

PR #219, opened 2019-09-05
chore(core-react): Add typings for checkbox

jamesdamild:
add types for checkbox

### DIFF
--- a/packages/core-react/index.d.ts
+++ b/packages/core-react/index.d.ts
@@ -37,9 +37,19 @@ export type ButtonProps = {
 
 declare const Button: React.FC<ButtonProps>;
 
+export type CheckboxProps = {
+  className?: string,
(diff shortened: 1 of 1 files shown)


packages/core-react/index.d.ts
```diff
@@ -37,9 +37,19 @@ export type ButtonProps = {
 
 declare const Button: React.FC<ButtonProps>;
 
+export type CheckboxProps = {
+  className?: string,
+  id?: string,
+  label?: React.ReactNode
+  name?: string,
```
seimithwework: What is `name` for again?

jamesdamild: Screen Shot 2019-09-05 at 1 21 12 PM
 for the checkbox, I based on the props that I see in index file for checkbox

PR #228, opened 2019-09-09
feat(core-react): add types for image

jgbenito7:
Adds typescript types for `Image` in `core-react`.

### DIFF
--- a/packages/core-react/index.d.ts
+++ b/packages/core-react/index.d.ts
@@ -93,14 +93,24 @@ export type SelectProps = {
 
 declare const Select: React.FC<SelectProps>;
 
-export { 
-  Button, 
+export type ImageProps = {
+  className?: string;
+  src: string;
+  aspect: '16by9' | '1by1' | '3by4' | '4by3' | 'cinema';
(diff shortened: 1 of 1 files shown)


packages/core-react/index.d.ts
```diff
@@ -93,14 +93,24 @@ export type SelectProps = {
 
 declare const Select: React.FC<SelectProps>;
 
-export { 
-  Button, 
+export type ImageProps = {
+  className?: string;
+  src: string;
```
jgbenito7: based on the source for Image, things will break if `src` or `aspect` are not supplied. That's why I have them as required.

packages/core-react/index.d.ts
```diff
@@ -93,14 +93,24 @@ export type SelectProps = {
 
 declare const Select: React.FC<SelectProps>;
 
-export { 
-  Button, 
+export type ImageProps = {
+  className?: string;
```
wuweiweiwu: I believe this is part of `HTMLProps<HTMLImageElement`

PR #229, opened 2019-09-11
James/selectprops functionality

jamesdamild:
**Ray Component**: https://ray.wework.com/components/select/

How to run locally
$ yarn storybook

Nutshell Details:
Semantic wrapper around Select  element:

`<Select />`

Screen Shot 2019-07-30 at 9 02 36 PM
Screen Shot 2019-07-30 at 9 02 45 PM
Screen Shot 2019-07-30 at 9 02 56 PM
Screen Shot 2019-07-30 at 9 03 08 PM

### DIFF
--- a/packages/core-react/index.d.ts
+++ b/packages/core-react/index.d.ts
@@ -80,6 +80,8 @@ export type CheckboxProps = {
 
 declare const Checkbox: React.FC<CheckboxProps>;
 
+declare type IconPositions = 'iconstart' | 'iconend';
+
 export type SelectProps = {
   className?: string;
   id?: string;
@@ -88,6 +90,11 @@ export type SelectProps = {
   compact?: boolean;
   disabled?: boolean;
   placeholder?: string;
+  prepend?: boolean,
+  icon?: React.ReactNode,
+  error?:boolean,
+  active?:boolean,
+  iconPosition?: IconPositions;
   children?: React.ReactNode;
 } & React.HTMLProps<HTMLSelectElement>;
 
--- a/packages/core-react/index.d.ts
+++ b/packages/core-react/index.d.ts
@@ -95,6 +95,7 @@ export type SelectProps = {
   error?:boolean,
   active?:boolean,
   iconPosition?: IconPositions;
+  onChange?:void;
   children?: React.ReactNode;
 } & React.HTMLProps<HTMLSelectElement>;
 
--- a/packages/core-react/index.d.ts
+++ b/packages/core-react/index.d.ts
@@ -95,7 +95,6 @@ export type SelectProps = {
   error?:boolean,
   active?:boolean,
   iconPosition?: IconPositions;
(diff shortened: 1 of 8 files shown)


packages/core-react/index.d.ts
```diff
@@ -88,6 +90,12 @@ export type SelectProps = {
   compact?: boolean;
   disabled?: boolean;
   placeholder?: string;
+  prepend?: boolean,
+  icon?: React.ReactNode,
+  error?:boolean,
+  active?:boolean,
+  iconPosition?: IconPositions;
+  onChange?:void;
```
seimithwework: Should we add `onFocus` and `onBlur` here as well?

packages/core-react/src/components/Select/Select.scss
```diff
@@ -0,0 +1,4 @@
+//overwriting ray componentdefault style -webkit-appearance: true
+.ray-select__input {
+  -webkit-appearance: none;
```
seimithwework: Do we need to also include the other things as well? 
```
-moz-appearance: ...
-o-appearance: ...
-ms-appearance: ...
appearance: ...
```

packages/core-react/src/components/Select/index.js
```diff
@@ -2,6 +2,13 @@ import React from 'react';
 import clsx from 'clsx';
 import PropTypes from 'prop-types';
 
+export const InputIcon = ({ prepend, icon }) => {
```
seimithwework: Nit: We can probably change the name of this component because it's not just for `Inputs` anymore. Also, feel free to move it out as a nested component. Otherwise, since it's a small thing here, I don't mind this being inside here too.

seimithwework:
Gotta write some tests! 😂

packages/core-react/src/components/Select/index.js
```diff
@@ -10,34 +17,88 @@ export default function Select({
...
+    setActiveState(true);
+  };
+
+  const handleBlur = () => {
+    setActiveState(false);
+  };
+
+  const handleChange = event => {
+    setValue(event.target.value);
+    onChange(event);
```
seimithwework: I'd write this defensively as `onChange && onChange(event);` so it won't throw an error when `onChange` isn't specified.

packages/core-react/src/components/Select/index.js
```diff
@@ -10,34 +17,88 @@ export default function Select({
...
+            className="ray-select__input"
+            name={name}
+            disabled={disabled}
+            onChange={handleChange}
+            onBlur={handleBlur}
+            onFocus={handleFocus}
+            value={currValue}
+            {...rest}
+          >
+            <option value="" disabled selected data-ray-placeholder>
```
seimithwework: For the `defaultValue` error i removed the `selected` attribute and that seemed to do the trick! 🤷‍♀

PR #232, opened 2019-09-16
feat(tabs): add / modify tabs component

gmharrison:
Adds tabs component

Desktop:
https://cl.ly/7a3cbd4d9991

Mobile:
https://cl.ly/55c0ea0ab4f0

### DIFF
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -1,0 +1,49 @@
+---
+label: Component
+title: Tabs
+---
+
+<page-intro>Tabs can be used to show content based on an individual tab's active state.</page-intro>
+
+<component
+    name="Tabs"
+    component="tabs"
+    variation="tabs"
+    >
+</component>
+
+<component
+    name="Tabs"
+    component="tabs"
+    variation="rtl-tabs"
+    >
+</component>
+
+# JavaScript API
+
+```javascript
+import { Tabs } from '@wework/ray-core';
+
+// instantiate all instances on document
+Tabs.createAll();
+
+// create instance
+const tabs = Tabs.create(document.querySelector('.ray-tabs'));
+
+// destroy
+tabs.destroy();
+```
+
+| Method                   | Params               | Description                      |
+| ------------------------ | -------------------- | -------------------------------- |
+| `Tabs.createAll`         | `HTMLElement:Object` | create all instances in document |
+| `Tabs.create`            | `HTMLElement:Object` | create an instance               |
+| `Tabs.prototype.destroy` |                      | destroy the instance             |
+
+## Modifiers
+
+Use these modifiers with `.ray-chip` class.
+
+| Selector                  | Description                             |
+| ------------------------- | --------------------------------------- |
+| `.ray-tabs__item--active` | Selector for applying active tab styles |
--- a/docs/components/tabs.md
+++ b/docs/components/tabs.md
@@ -44,6 +44,7 @@ tabs.destroy();
 
 Use these modifiers with `.ray-chip` class.
 
-| Selector                  | Description                             |
-| ------------------------- | --------------------------------------- |
-| `.ray-tabs__item--active` | Selector for applying active tab styles |
+| Selector                     | Description                                 |
+| ---------------------------- | ------------------------------------------- |
+| `.ray-tabs__item--active`    | Selector for applying active tab styles     |
+| `.ray-tabs__content--active` | Selector for applying active content styles |
--- a/docs/html/tabs/tabs.html
+++ b/docs/html/tabs/tabs.html
@@ -1,5 +1,29 @@
-<ul class="ray-tabs">
-  <li tabindex="1" class="ray-tabs__item">Desktop</li>
-  <li tabindex="1" class="ray-tabs__item">Tablet</li>
-  <li tabindex="1" class="ray-tabs__item ray-tabs__item--active">Mobile</li>
-</ul>
+<div class="ray-tabs">
+  <ul class="ray-tabs__list">
+    <li tabindex="1" data-tab-order="1" class="ray-tabs__item">
+      Desktop
+    </li>
+    <li tabindex="1" data-tab-order="2" class="ray-tabs__item">
+      Tablet
+    </li>
+    <li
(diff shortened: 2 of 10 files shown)


docs/components/tabs.md
```diff
@@ -0,0 +1,50 @@
...
+
+| Method                   | Params               | Description                      |
+| ------------------------ | -------------------- | -------------------------------- |
+| `Tabs.createAll`         | `HTMLElement:Object` | create all instances in document |
+| `Tabs.create`            | `HTMLElement:Object` | create an instance               |
+| `Tabs.prototype.destroy` |                      | destroy the instance             |
+
+## Modifiers
+
+Use these modifiers with `.ray-chip` class.
```
adamraider:
```suggestion
Use these modifiers with `.ray-tabs` class.
```

packages/core/src/components/tabs/_tabs.scss
```diff
@@ -6,28 +6,50 @@ $tabs-border-size: 1px;
...
       color: $ray-color-blue-50;
       text-align: center;
-      min-width: 200px;
-      padding-top: 1rem;
-      padding-bottom: 1rem;
       margin-bottom: -$tabs-border-size !important;
+      min-width: 85px;
+      outline: 0;
+
+      @media (min-width: map-get($ray-layout-grid-breakpoints, tablet)) {
```
adamraider: prefer to use our mixins they're much more readable 
```suggestion
      @include ray-breakpoint(tablet)
```

packages/core/src/components/tabs/index.js
```diff
@@ -0,0 +1,104 @@
...
+    this.contentItems = Array.from(
+      this._root.querySelectorAll('.ray-tabs__content')
+    );
+
+    this._bindEventListeners();
+  }
+
+  _bindEventListeners() {
+    this.tabItems.forEach(tab => {
+      tab.addEventListener('mousedown', this.onClick);
```
adamraider: will this work on mobile?

packages/core/src/components/tabs/index.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  constructor(root, options) {
+    this._root = root;
+    this._options = options;
+    this.constructor.instances.set(this._root, this);
+
+    this.tabItems = Array.from(this._root.querySelectorAll('.ray-tabs__item'));
+    this.contentItems = Array.from(
+      this._root.querySelectorAll('.ray-tabs__content')
```
adamraider: i would suggest using a data- identifier for the content and subsequently querying the entire dom for the content portion. This way if we want to use tabs but have the content appear in a ancestor element it will be possible

packages/core/src/components/tabs/_tabs.scss
```diff
@@ -6,28 +6,50 @@ $tabs-border-size: 1px;
 
 @include exports('ray-tabs') {
   .#{$ray-class-prefix}tabs {
-    border-bottom: $tabs-border-size solid $ray-color-blue-50;
-    display: flex;
+    &__list {
+      border-bottom: $tabs-border-size solid $ray-color-gray-10;
+      display: flex;
```
adamraider: we should probably wrap these elements, we don't want the ui to break if they start overflowing

packages/core/src/components/tabs/index.js
```diff
@@ -0,0 +1,104 @@
...
+
+  assignTabClasses(event) {
+    this.tabItems.forEach(tab => {
+      tab.classList.remove(this.constructor.cssClasses.ACTIVE);
+    });
+    event.currentTarget.classList.add(this.constructor.cssClasses.ACTIVE);
+  }
+
+  assignTabContentClasses(event) {
+    const tabNumber = event.currentTarget.getAttribute('data-tab-order');
```
adamraider: the name of this data- attr is a little confusing to me, why not call it something like  `data-tab-id` ?

packages/core/src/components/tabs/_tabs.scss
```diff
@@ -6,28 +6,50 @@ $tabs-border-size: 1px;
...
+      padding-bottom: 0.75rem;
       display: block;
       cursor: pointer;
       color: $ray-color-blue-50;
       text-align: center;
-      min-width: 200px;
-      padding-top: 1rem;
-      padding-bottom: 1rem;
       margin-bottom: -$tabs-border-size !important;
+      min-width: 85px;
```
adamraider: image

this solution doesn't work that well with labels >85px , as there then wont be any spacing between them

```suggestion
    padding-left: 0.5rem;
    padding-right: 0.5rem;
```

packages/core/src/components/tabs/_tabs.scss
```diff
@@ -6,28 +6,50 @@ $tabs-border-size: 1px;
...
+      outline: 0;
+
+      @media (min-width: map-get($ray-layout-grid-breakpoints, tablet)) {
+        font-size: 1.25rem;
+        min-width: 200px;
```
adamraider: same comment as above

adamraider (changes requested):
Really cool! We've been needing these for a while.

I made a few suggestions

docs/html/tabs/tabs.html
```diff
@@ -0,0 +1,29 @@
+<div class="ray-tabs">
```
adamraider: consider adding the proper aria attrs to the example

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role

so when people copy/paste they can just edit them in super easily

adamraider: https://material.io/develop/web/components/tabs/tab-bar/ use material as a reference

packages/core/src/components/tabs/_tabs.scss
i8ramin: could use `text-overflow` -- https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

seimithwework:
How are we dealing with accessibly by chance? Also, do we want the interactivity here and not in `core-react`?

PR #233, opened 2019-09-23
feat(grid): add product grid to ray product

Morgantheplant:
https://jira.weworkers.io/browse/WEB-754

Moves the Container, Grid, and Cell over from [core-components](https://github.com/WeConnect/wework-web-monorepo/tree/master/packages/core-components). 

I also updated the jest config in ray product-react to also check `.tsx`, `.ts`, and `.jsx` extensions. It looks like the tests weren't running on the tsx files. 

Packages added:
**enzyme-to-json** - for snapshots to run similar to core-components
**identity-obj-proxy** - for stubbing out css modules in tests

The code here was mostly untouched except for swapping out `classnames` package for `clsx`


Morgantheplant:
added `@/preset-typescript` to root bc I was failing when ran from root (not from the ray-product directory). Then it was failing bc of scss import error so I added `identity-obj-proxy` to root as well. Tried removing it from the ray-product directory but that would fail locally, so added it back in there.

packages/product-react/src/components/Container/Container.test.tsx
```diff
@@ -0,0 +1,147 @@
+import Enzyme, { configure, mount, ReactWrapper } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import * as React from 'react';
+import { Container } from './Container';
+
+configure({ adapter: new Adapter() });
```
wuweiweiwu: this should be handled in `test/setup.ts`?

packages/product-react/stories/Grid.story.tsx
```diff
@@ -0,0 +1,310 @@
+import { storiesOf } from '@storybook/react';
+import * as React from 'react';
+import { Container } from '../src/components/Container';
+import '../src/components/Container/Container.scss'; // TO DO THIS IS A WEIRD PLACE
```
wuweiweiwu: I believe you can put this in `configureStories`

jest.config.js
```diff
@@ -8,5 +8,17 @@ module.exports = {
       displayName: '@wework/rehype-react',
       testMatch: ['<rootDir>/packages/rehype-react/**/*.test.js']
     }
-  ]
+  ],
```
wuweiweiwu: do these changes will need to exist at toplevel?

packages/product-react/src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,29 @@
...
+import Cell from './Cell';
+
+type GridProps = {
+  as?: keyof ReactHTML;
+  children?: ReactNode;
+  className?: string;
+} & typeof defaultProps;
+
+export const Grid = ({
+  as: ComponentType,
```
wuweiweiwu: might be out of scope for this PR but in the future I would like to see us do something like this 

https://github.com/reakit/reakit/blob/master/packages/reakit-system/src/createComponent.ts#L21

where you can infer the props of the `as` component and assert them

wuweiweiwu:
and the CI issue is because when you're referencing `@wework/ray-product-react` in the test it isn't built yet.

https://github.com/wework/ray/blob/master/packages/product-react/package.json#L5

packages/product-react/src/components/Container/Container.scss
```diff
@@ -0,0 +1,78 @@
+$lucia-container-margin-1: 0 12px;
```
wuweiweiwu: now that we're distributing new styles, this needs to be part of the build.

we need to output the compiled scss as part of `dist`

packages/product-react/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,68 @@
+import clsx from 'clsx';
+import React, { ReactHTML, ReactNode } from 'react';
+
+export type ContainerProps = {
+  align?: 'left' | 'right';
```
seimithwework: Can we update this and the other string literal type (`justify`) to something like this?
```
export enum AlignNameHere {
  LEFT = 'small',
  RIGHT = 'right',
}
...
align?: keyof typeof AlignNameHere;
```

packages/product-react/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,68 @@
...
+  children?: ReactNode;
+  className?: string;
+  debug?: boolean;
+  fixed?: boolean;
+  justify?: 'left' | 'center' | 'right';
+} & typeof defaultProps;
+
+const containerJustifyClass = (justify?: string) => {
+  switch (justify) {
+    case 'left':
```
seimithwework: We probably have to update these to not be hardcoded if we do the enums.

packages/product-react/src/components/Grid/Cell.tsx
```diff
@@ -0,0 +1,76 @@
+import clsx from 'clsx';
+import React, { ReactHTML, ReactNode } from 'react';
+
+export type CellProps = {
+  align?: 'top' | 'middle' | 'bottom';
```
seimithwework: Same comment I had earlier :-P

packages/product-react/stories/Cell.story.tsx
```diff
@@ -0,0 +1,69 @@
+import { storiesOf } from '@storybook/react';
+import * as React from 'react';
+import { Grid } from '../src/components/Grid';
+import '../src/components/Container/Container.scss';
+
+storiesOf('Core Comps | Cell', module)
```
seimithwework: This is a nit, but i think we can remove mentions of "core components" when we move stuff into Ray.

packages/product-react/src/components/Container/Container.scss
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/* Lucia Debug ( debug={true} ) */
+.ray-page-container--debug {
+  &.ray-page-container {
+    background: yellow;
+  }
+
+  .ray-grid__cell {
+    background: hotpink;
```
seimithwework: `yellow` `hotpink` lol

Morgantheplant:
closing in favor of https://github.com/wework/ray/pull/234

PR #235, opened 2019-10-02
feat(core-react): change type to kind on Button

i8ramin:
The `type` attribute is a valid Button attribute (can have values `submit`, `reset` and `button` - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), so changing the current `type` attribute to `kind`


packages/core-react/src/components/Button/index.js
```diff
@@ -45,7 +45,7 @@ Button.propTypes = {
 
 Button.defaultProps = {
   Tag: 'button',
-  type: 'primary',
+  kind: 'primary',
```
seimithwework: I understand how this conflicts with HTML but I'm not digging the `kind` name. Can we do variant? Also, this would be a breaking change right? It looks like 4 other teams/repos are importing ray (but i guess they may/may not be using the React implementation of it). Otherwise, I'm OK with this update, but I'd want to hear from other folks about the name of this prop. (@jamesdamild would love your feedback as well :-P)

i8ramin: `variant` works as well. not issues there. just can't be `type`

wuweiweiwu: +1 on variant

fwiw material ui uses `variant` and semantic ui react has individual boolean props for each of the variants

wuweiweiwu:
+1 on the updating other components.

packages/core-react/package.json
```diff
@@ -9,21 +9,26 @@
     "dev": "yarn storybook",
     "build": "gulp scripts:es scripts:umd scripts:compiled",
     "test": "jest",
+    "tdd": "jest --watch",
```
adamraider:
```suggestion
    "test:watch": "jest --watch",
```

maybe a bit more clear, not really sure if this is necessary

packages/core-react/package.json
```diff
@@ -66,8 +71,7 @@
     "README.md",
     "index.d.ts"
   ],
-  "peerDependencies": {
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6"
+  "jest": {
```
adamraider: this should go in the jest.configs

adamraider: we want to avoid multiple sources of configuration

packages/core-react/package.json
i8ramin: i personally prefer typing less :-P

i8ramin:
Closing in favor of https://github.com/wework/ray/pull/241

PR #237, opened 2019-10-08
converted checkbox from js to tsx in ray-product

jamesdamild:
Ray Component: https://ray.wework.com/components/checkbox/

How to run locally
$ yarn storybook

Nutshell Details:
Semantic wrapper around Checkbox element:

<Checkbox />

Screen Shot 2019-10-08 at 2 01 22 PM

### DIFF
--- a/packages/product-react/src/components/Checkbox/Checkbox.test.tsx
+++ b/packages/product-react/src/components/Checkbox/Checkbox.test.tsx
@@ -7,6 +7,7 @@ describe('Checkbox', () => {
   test('it renders a checkbox', () => {
     const wrapper = Enzyme.mount(<Checkbox label="hello world" />);
     expect(wrapper.find('.ray-checkbox').length).toBe(1);
+    expect(wrapper.find('input[type="checkbox"]').length).toBe(1);
   });
 
   test('it supports a custom class', () => {
--- a/packages/product-react/src/components/Checkbox/Checkbox.test.tsx
+++ b/packages/product-react/src/components/Checkbox/Checkbox.test.tsx
@@ -1,0 +1,18 @@
+import Enzyme from 'enzyme';
+import React from 'react';
+
+import Checkbox from './Checkbox';
+
+describe('Checkbox', () => {
+  test('it renders a checkbox', () => {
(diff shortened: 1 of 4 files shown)


packages/product-react/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,18 @@
+import Enzyme from 'enzyme';
+import React from 'react';
+
+import Checkbox from './Checkbox';
+
+describe('Checkbox', () => {
+  test('it renders a checkbox', () => {
+    const wrapper = Enzyme.mount(<Checkbox label="hello world" />);
+    expect(wrapper.find('.ray-checkbox').length).toBe(1);
```
Morgantheplant: nit this test doesn't actually assert if there's a checkbox input

packages/product-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <div className={clsx('ray-checkbox', className)}>
+      <input
+        className="ray-checkbox__input"
+        type="checkbox"
+        id={id}
+        name={name}
+        {...rest}
+      />
+      <label className="ray-checkbox__label" htmlFor={id}>
```
Morgantheplant: i wonder if you should make the id required here to prevent that htmlFor from being `undefined`. You could also  you can nest the `<input>` directly inside the `<label>` and the htmlFor and id attributes are not needed because the association is implicit
```
<label><input/>{label}</label>
```

packages/product-react/src/components/Checkbox/Checkbox.test.tsx
jamesdamild: done

packages/product-react/src/components/Checkbox/Checkbox.tsx
jamesdamild: ya good point for the required id , the way the may the made it in the ray CSS ,the focus stop working if you don't put the htmlFor as an attribute. really weird :/

PR #238, opened 2019-10-08
James/ray product/input js to tsx

jamesdamild:
Ray Component: https://ray.wework.com/components/text-field/

How to run locally
$ yarn storybook

Nutshell Details:
Semantic wrapper around Input element:

<Input/>

Screen Shot 2019-10-08 at 5 19 02 PM
Screen Shot 2019-10-08 at 5 19 12 PM
Screen Shot 2019-10-08 at 5 19 22 PM
Screen Shot 2019-10-08 at 5 19 30 PM
Screen Shot 2019-10-08 at 5 19 41 PM
Screen Shot 2019-10-08 at 5 19 49 PM
Screen Shot 2019-10-08 at 5 19 57 PM


package.json
```diff
@@ -37,12 +37,16 @@
     "eslint-plugin-import": "2.17.2",
     "eslint-plugin-prettier": "3.0.1",
     "eslint-plugin-react": "7.13.0",
-    "husky": "2.3.0",
+    "husky": "^3.0.8",
```
jgbenito7: Could you add a comment here to explain why this was updated?

packages/product-react/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,112 @@
...
+    'ray-text-field--error': error,
+    'ray-text-field--has-value': value,
+    'ray-text-field--active': active || focus,
+    'ray-text-field--with-icon-start':
+      iconPosition === IconPosition.Start && icon,
+    'ray-text-field--with-icon-end': iconPosition === IconPosition.End && icon,
+    'ray-text-field--with-prepend':
+      iconPosition === IconPosition.Prepend && icon
+  });
+  const handleFocus = (e: React.FocusEvent<HTMLInputElement>) => {
```
jgbenito7: nit: you could reuse this type

packages/product-react/src/components/Input/Input.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+  };
+
+  beforeEach(() => {
+    component = Enzyme.mount(<Input {...defaultProps} />);
+  });
+
+  test('it renders a input', () => {
+    const wrapper = component;
+    expect(wrapper.find('.ray-text-field').length).toBe(1);
```
Morgantheplant: nit this test doesn't look for an input element

packages/product-react/src/components/Input/Input.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    it('has value property that can be set (:number || string)', () => {
+      const component = Enzyme.mount(<Input value={4} {...defaultProps} />);
+      expect(component.props().value).toEqual(4);
+    });
+  });
+
+  describe('Functions events', () => {
+    it('onFocus', () => {
+      component.setProps({ onFocus: fauxCallback });
+      component.find('input').simulate('focus');
```
Morgantheplant: it looks like these tests don't assert anything. Also the description is not descriptive.  maybe something more like `should call 'onFocus' prop on input focus`

packages/product-react/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,112 @@
...
+import React from 'react';
+import FieldWrapper from '../Common/FieldWrapper';
+import FormItemWrapper from '../Common/FormItemWrapper';
+import IconWrapper from '../Common/IconWrapper';
+import RTLWrapper from '../Common/RTLWrapper';
+
+export enum IconPosition {
+  Prepend = 'PREPEND',
+  End = 'END',
+  Start = 'START'
```
Morgantheplant: nit these should be all UPPERCASE https://github.com/airbnb/javascript#naming--uppercase

packages/product-react/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return (
+    <RTLWrapper renderWrapper={rtl}>
+      <FormItemWrapper renderWrapper={formItem}>
+        <div className={wrapperClass}>
+          <IconWrapper
+            renderWrapper={iconPosition === IconPosition.Prepend}
+            iconClass="ray-text-field__prepend"
+          >
+            {icon}
+          </IconWrapper>
```
Morgantheplant: should this be? 
```
{icon && <IconWrapper
            renderWrapper={iconPosition === IconPosition.Prepend}
            iconClass="ray-text-field__prepend"
          >
            {icon}
          </IconWrapper>}
```

packages/product-react/src/components/Input/Input.tsx
```diff
@@ -0,0 +1,112 @@
...
+              {...rest}
+            />
+            {label && (
+              <label htmlFor={id} className="ray-text-field__label">
+                {label}
+              </label>
+            )}
+          </FieldWrapper>
+        </div>
+        {(formItem || hint) && (
```
Morgantheplant: the logic here seems interesting. Is there a scenario where `formItem` would be true and hint missing? if so would you want to render the hint?

jamesdamild: Uhmm good question didn't  write this one, was copied over from the monorepo , if you look on the line 148,input.tsx inside the monorepo you'll see it

packages/product-react/src/components/Input/Input.tsx
jamesdamild: uhmm good to know, but was following this convention, https://www.typescriptlang.org/docs/handbook/enums.html#string-enums, what do you think ?

packages/product-react/src/components/Input/Input.tsx
jamesdamild: oo ok good to know ,  didn't get it, can you provide an example so I can see what you meant, please?

package.json
jamesdamild: oo didn't see that, I put it back the  olf version

packages/product-react/src/components/Input/Input.tsx
jamesdamild: good

packages/product-react/src/components/Input/Input.test.tsx
jamesdamild: donr

package.json
jamesdamild: done

packages/product-react/src/components/Input/Input.tsx
Morgantheplant: oo that's interesting. I've always used uppercase in both JS and python. a nit though so up to you!

packages/product-react/src/components/Input/Input.tsx
wuweiweiwu: as the original author:

if the hint exist we always want to render the formItem wrapper

so this logic would probably be just `hint && (...` here and on line71 would be `hint || formItem`

But I would prefer `FormItem` to be a separate component as it is a superset of `TextField`

tslint.json
```diff
@@ -0,0 +1,34 @@
+{
```
wuweiweiwu: where are we getting these configurations from?

package.json
```diff
@@ -37,12 +37,16 @@
...
     "eslint-plugin-react": "7.13.0",
-    "husky": "2.3.0",
+    "husky": "^2.3.0",
     "jest": "24.8.0",
     "lerna": "3.13.4",
     "lint-staged": "8.1.4",
     "prettier": "1.16.4",
-    "stylelint": "^9.10.1"
+    "stylelint": "^9.10.1",
+    "tslint": "^5.11.0",
```
wuweiweiwu: why don't we use eslint for this?

https://github.com/palantir/tslint/issues/4534

we use eslint to lint all js and ts files in grant

packages/product-react/src/components/Select/Select.test.tsx
```diff
@@ -119,16 +119,19 @@ describe('Select', () => {
...
       component.setProps({ onBlur: fauxCallback });
       component.find('select').simulate('blur');
+      expect(component.find('.ray-select--active').exists()).toBe(false);
     });
 
     it('onChange', () => {
-      component.setProps({ onChange: fauxCallback });
+      component.setProps({ onChange: fauxCallback, value: 4 });
       component.find('select').simulate('change');
+      expect(fauxCallback).toHaveBeenCalled();
```
wuweiweiwu: if we're adding a value on L132 shouldn't we be asserting that the callback is called with `4`?

package.json
jamesdamild: nice thanks for the tip, i removed tslint

packages/product-react/src/components/Select/Select.tsx
```diff
@@ -8,9 +8,9 @@ import RTLWrapper from '../Common/RTLWrapper';
 import './Select.scss';
 
 export enum IconPosition {
-  Prepend = 'PREPEND',
-  End = 'END',
-  Start = 'START'
+  PREPEND = 'PREPEND',
```
adamraider: What's the purpose of this change?

package.json
jamesdamild: no, it doesn't hurt  ,  works fine

packages/product-react/src/components/Select/Select.tsx
jamesdamild: This to add the prepend icon for the select component, just in case the user wants it
Screen Shot 2019-11-01 at 1 38 00 PM

PR #241, opened 2019-10-14
feat(core-react): change "type" to "variant" on Button

i8ramin:
The `type` attribute is a valid Button attribute (can have values `submit`, `reset` and `button` - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), so changing the current `type` attribute to `variant`

original PR: https://github.com/wework/ray/pull/235


packages/core-react/package.json
```diff
@@ -9,21 +9,26 @@
     "dev": "yarn storybook",
     "build": "gulp scripts:es scripts:umd scripts:compiled",
     "test": "jest",
+    "tdd": "jest --watch",
```
wuweiweiwu: nit: `test:watch`

i8ramin: I'm the only one who prefers `tdd`, huh?

wuweiweiwu: personally I usually pass the flags in the yarn commands not super opinionated about this though

PR #243, opened 2019-10-22
Feat(ray-product):Added CARD in tsx format 

jamesdamild:
Ray Component: https://ray.wework.com/components/card/

How to run locally
$ yarn storybook

Nutshell Details:
Semantic wrapper around Card element:


packages/product-react/src/components/Card/Card.scss
```diff
@@ -0,0 +1,15 @@
+.lucia-card--compact {
+  padding: 1rem;
```
seimithwework: We can variable-ize these values; Also, we should change the `lucia` name to something else lol

seimithwework: We can probably dry these up by nesting `&--compact` and `&--inactive` under `.lucia` (after we change it to some other name :-P)

packages/product-react/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,92 @@
...
+  compact,
+  inactive,
+  children,
+  className,
+  rtl,
+  linkProps,
+  ...rest
+}) => {
+   return (
+    <LINKWrapper props={linkProps}>
```
seimithwework: Hmmm these names are interesting! I think regular Capitalized camelcase should be ok instead.

packages/product-react/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,92 @@
...
+              'ray-card--row': row,
+              'lucia-card--compact': compact,
+              'lucia-card--inactive': inactive
+            },
+            className
+          )}
+          {...rest}
+        >
+          {image && (
+            <div className="ray-card__image ray-image ray-image--16by9">
```
jgbenito7: These 16x9 classes always made me feel uncomfortable..

packages/product-react/src/components/Card/Card.scss
jamesdamild: yes agree, but I have no control over his name, maybe it can be changed in another pull request for the project where it's defined?

PR #244, opened 2019-10-24
Feat(ray-product):Added Image in tsx format 

jamesdamild:
Ray Component: https://ray.wework.com/components/image/

How to run locally
$ yarn storybook
$ yarn test  product-react  --watch

Nutshell Details:
Semantic wrapper around Image element:

### DIFF
--- a/packages/product-react/src/components/Image/Image.tsx
+++ b/packages/product-react/src/components/Image/Image.tsx
@@ -17,6 +17,7 @@ export interface IImageProps extends React.HTMLProps<HTMLImageElement> {
   useMap?: string;
   caption?: string;
   decoding?: 'async' | 'auto' | 'sync';
+  style?: React.CSSProperties;
 }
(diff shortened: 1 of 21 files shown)


packages/product-react/src/components/Image/Image.tsx
```diff
@@ -0,0 +1,82 @@
...
+        className={clsx(
+          'ray-bg',
+          {
+            [`ray-bg--${aspect}`]: aspect
+          },
+          className
+        )}
+        {...rest}
+        style={{
+          ...rest.style,
```
Morgantheplant: Seems like you should  pull style out via destructing above. That was you can just do `...style` And you don’t set this property and then overwrite it.

jamesdamild: Ya good catch. I fixed it thank bud

PR #246, opened 2019-10-28
feat(core): Add support for append addon which works like prepend one

sick-sad-world:
Proposal to add "append" addon, that works like existing "prepend" one. In order to avoid of modifying ray components in `www-website` project. For examples see docs/storybook

### DIFF
--- a/packages/core/stories/text-field.stories.js
+++ b/packages/core/stories/text-field.stories.js
@@ -760,7 +760,7 @@ storiesOf('Text Field', module)
                   required
                 />
                 <label className="ray-text-field__label" htmlFor="input-email">
-                  Email Adress
+                  Email Address
                 </label>
               </div>
             </div>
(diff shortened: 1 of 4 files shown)


packages/core/stories/text-field.stories.js
```diff
@@ -671,6 +671,270 @@ storiesOf('Text Field', module)
...
+              <div className="ray-text-field__wrapper">
+                <input
+                  type="text"
+                  className="ray-text-field__input"
+                  id="input-email"
+                  placeholder="john@example.com"
+                  required
+                />
+                <label className="ray-text-field__label" htmlFor="input-email">
+                  Email Adress
```
primeare: Mistype:
```suggestion
                  Email Address
```

packages/core/stories/text-field.stories.js
```diff
@@ -671,6 +671,270 @@ storiesOf('Text Field', module)
...
```
primeare: Mistype:
```suggestion
                  Email Address
```

PR #251, opened 2019-11-11
Feat: Added avatar to product-react

jamesdamild:
How to run locally
$ yarn storybook

Props Definitions:
  add?: boolean;
  alt?: string;
  edit?: boolean;
  name?: string;
  image?: string;
  size?:  'small' or  'medium' or  'large'
  className?: string;
  onAddImageClick?: (event?: React.SyntheticEvent<EventTarget>) => void
 ..rest


packages/product-react/src/components/Avatar/AvatarIcons.tsx
```diff
@@ -0,0 +1,53 @@
...
+        id="Rectangle-path"
+        fill="#FFFFFF"
+        fillRule="nonzero"
+        x="7"
+        y="21"
+        width="13"
+        height="1"
+      />
+      <rect
+        id="Rectangle-path"
```
Morgantheplant: we should prob strip out all the IDs in this SVG

PR #253, opened 2019-11-13
  Feat (Product-react):List Component

jamesdamild:
List Component
JIRA Ticket: [WEB-518]( https://jira.weworkers.io/browse/WEB-518)

**Change Details**
   Created a List component

**Props**
  items: Array of objects for items[{
  header?, 
  content?,
  icon?
 }];
 
listItemComponent?:Optional Prop to pass custom component for list item.

Example on how to create a custom component to pass as list items:
First, you have to import interfaces IListProps and items from the list component it cant be :

import { List, IListProps, IItems } from '../src/components/List/List';

***1) Your custom component has to be the type of IListProps***
***2) When looping through the items, item has to be the type of IItems***

and that's it you're good to go

***If you don't pass a listItemComponent, the default one will be the avatar list item.***

Custom List item component:
```
const customListItem: React.FC<IListProps> = ({
  items,
  className,
  ...rest
}) => {
  let listItems: React.ReactNode[];
  listItems = items.map((item: IItems, index: number) => {
    return (
      <span>
        <p>
          <Checkbox id="1" style={{ display: 'inline-block' }} />
        </p>
        <p
          style={{
            marginLeft: '38px',
            marginBottom: '17px',
            marginTop: '-12px'
          }}
        >
          <span className="ray-list-item__header">{item.header}</span>
          <span className="ray-list-item__content">{item.content}</span>
        </p>
      </span>
    );
  });
  return <ul>{listItems}</ul>;
};


Example  of default component  with avatar is created :

const fauxAvatarImage =
  'https://pbs.twimg.com/profile_images/624249118114881536/qxn_I_oR_400x400.jpg';

 <List
        items={[
          {
            header: 'Sammy Lawson',
            content: 'Product Designer',
            icon: fauxAvatarImage
          },
          {
            header: 'Arnold Schwarzenegger',
            content: 'Product Designer',
            icon: fauxAvatarImage
          },
          {
            header: 'Michael Jordan',
            content: 'Product Designer',
            icon: fauxAvatarImage
          }
        ]}
      />

```


packages/product-react/src/components/List/List.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+              content: 'Product Designer',
+              icon: 'laalalalaaal',
+            },
+            {
+              header: 'Michael Jordan',
+              content: 'Product Designer',
+              icon: 'laalalalaaallaalalalaaal',
+            },
+          ]}
+        />,
```
Morgantheplant: It looks like this test doesn't assert anything

packages/product-react/src/components/List/List.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const ListItem: React.FC<IListProps> = ({ items, className, ...rest }) => {
+  let listItems: React.ReactNode[];
+  listItems = items.map((item: IItems, index: number) => {
+    return (
+      <li className={className} {...rest} key={index}>
+        {item.icon && (
+          <Avatar
+            className="icon-block"
+            image={String(item.icon)}
```
Morgantheplant: Will this ever not be a string? Could that instead be enforced via typescript?

Morgantheplant: what about the fallback `name` property for Avatar?

packages/product-react/src/components/List/List.tsx
```diff
@@ -0,0 +1,58 @@
...
+            size="medium"
+          />
+        )}
+        <p className={'container'}>
```
Morgantheplant: the brackets here are redundant

packages/product-react/stories/List.story.tsx
```diff
@@ -0,0 +1,111 @@
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import { List, IListProps, IItems } from '../src/components/List/List';
+import { Checkbox } from '../src/components/Checkbox';
+import '../src/components/List/List.scss';
+
+const customListItem: React.FC<IListProps> = ({
+  items,
+  className,
+  ...rest
```
Morgantheplant: looks like this can be removed

Morgantheplant: wait this is a story so it doesn't matter too much

jamesdamild: yes it is, it's just to demo a custom component

packages/product-react/src/components/List/List.tsx
jamesdamild: yes your so right lol

packages/product-react/src/components/List/List.test.tsx
jamesdamild: oo   your right and also don't even need this  test, there is no focus my bad

packages/product-react/src/components/List/List.tsx
jamesdamild: uhmm good point let me investigate with @seimithwework

packages/product-react/src/components/List/List.scss
```diff
@@ -0,0 +1,51 @@
+@import url('https://www-static.wework.com/apercu/apercu.css');
+@import url('https://www-static.wework.com/apercu/apercu_mono.css');
+
+$list-font-family: 'Apercu';
```
seimithwework: Do we need to set the font-family here? I was thinking that it would inherit it from somewhere else. But, if we have to set it, we should probably have some fallback fonts. 

I just grabbed this from wework.com haha:
`'Apercu-Regular', 'Apercu', Avenir, -apple-system, BlinkMacSystemFont, 'Helvetica Neue', Helvetica, 'Calibri', 'Roboto', Arial, sans-serif`

packages/product-react/src/components/List/List.scss
```diff
@@ -0,0 +1,51 @@
...
+$list-font-weight: normal;
+$list-block-display: block;
+$list-inline-block-display: inline-block;
+$list-vertical-align: middle;
+
+.ray-list-item {
+  font-family: $list-font-family;
+  font-weight: $list-font-weight;
+  border-radius: 4px;
```
seimithwework: Variable-ize these please 🙏!

packages/product-react/src/components/List/List.scss
```diff
@@ -0,0 +1,51 @@
...
+  background-color: rgb(247, 247, 247);
+}
+
+.ray-list-item__header {
```
seimithwework: 😂I'm sure I'm the culprit for these not being nested!! lol Sorry! but we can probably nest these like so:
```
.ray-list-item {
    &__header {
        ...
    }
}
```

packages/product-react/src/components/List/List.scss
jamesdamild: those vaues come from simon lol https://app.goabstract.com/projects/48da46b0-6077-11e9-8ae4-61e174908c7b/branches/master/collections/d2738c26-c962-4ab3-8679-127fcbc77cc1?collectionLayerId=8a6df6f3-aea5-4eb4-aba2-d30de739b6e3&mode=build&selected=1867306355-084D6094-A393-44C6-91F9-E5FCE5045AEF

packages/product-react/src/components/List/List.scss
jamesdamild: ok done

packages/product-react/src/components/List/List.scss
jamesdamild: done

packages/product-react/src/components/List/List.tsx
seimithwework: Oh yeah, we can set the `image` type to string. What do you by fallback `name` property? Are you talking about when you pass in a "name" to the `Avatar` and we grab the first initial and display that? @Morgantheplant

jamesdamild: > Will this ever not be a string? Could that instead be enforced via typescript?

yes it's already enforced, inside the avatar image is the type of string, I think the image will always be base64, so if pass a base64, should work also. will remove the string part and will type it of string

PR #255, opened 2019-11-28
Event propagation fix for Dropdown component

sick-sad-world:
Fixed event propagation for onClick handler and provided correct target for option click handler

### DIFF
--- a/packages/core/src/components/dropdown/index.js
+++ b/packages/core/src/components/dropdown/index.js
@@ -358,7 +358,7 @@ class Dropdown {
   };
 
   onOptionClick(plugin) {
-    return function onClickListener() {
+    return function onClickListener(e) {
       if (this.hasAttribute('disabled') || !this.dataset.rayIdx) return;
       plugin._value = plugin._options[this.dataset.rayIdx].value; //eslint-disable-line
     };
--- a/packages/core/src/components/dropdown/index.js
+++ b/packages/core/src/components/dropdown/index.js
@@ -358,7 +358,7 @@ class Dropdown {
   };
 
   onOptionClick(plugin) {
-    return function onClickListener(e) {
+    return function onClickListener() {
       if (this.hasAttribute('disabled') || !this.dataset.rayIdx) return;
       plugin._value = plugin._options[this.dataset.rayIdx].value; //eslint-disable-line
     };
--- a/packages/core/src/components/dropdown/index.js
+++ b/packages/core/src/components/dropdown/index.js
@@ -311,7 +311,21 @@ class Dropdown {
     switchClassName(this._root, ['ACTIVE', 'OPEN'], true);
   };
 
-  update = () => {
+  update = options => {
+    if (options) {
+      this.settings = {
+        ...defaults,
+        ...options
+      };
+    }
+    markupTemplates.forEach(template => {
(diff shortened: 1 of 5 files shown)


cmugla:
I'm a bit confused what the bug with propagation was, can you provide a description of the bug you were seeing in the description?

Also, please add a test!

packages/core/src/components/dropdown/index.js
```diff
@@ -308,7 +308,21 @@ class Dropdown {
...
 
-  update = () => {
+  update = options => {
+    if (options) {
+      this.settings = {
+        ...defaults,
+        ...options
+      };
+    }
+    markupTemplates.forEach(template => {
```
cmugla: is this copy/pasted from the constructor? If so, let's make it its own function.

packages/core/src/components/dropdown/index.js
```diff
@@ -330,8 +344,6 @@ class Dropdown {
   };
 
   onClick = e => {
-    e.stopPropagation();
-    e.preventDefault();
```
cmugla: why are these removed?

cmugla:
Just merged the fix for the deploy-preview - please pull master and push back up. Hopefully will do the trick

packages/core/src/components/dropdown/index.js
sick-sad-world: done

packages/core/src/components/dropdown/index.js
sick-sad-world: this is removed because click handler is listening entire body to handle "click outside" and this is removed that it would not interfere with other click listeners. For example form submit button e.t.c Found this issue while integrating Dropdown into Landlords form

packages/core/src/components/dropdown/index.js
```diff
@@ -315,15 +317,9 @@ class Dropdown {
         ...options
       };
     }
-    markupTemplates.forEach(template => {
```
cmugla: 🤔 this was removed, is it no longer needed or should you be calling `this._renderMarkup`?

packages/core/src/components/dropdown/index.js
cmugla: ah ok

cmugla:
Also want to note here that the Error state looks weird:
- the error hint is nested within the dropdown
- the dropdown is cut off by the message
- the border color is not the same when open

Image 2019-12-04 at 3 53 54 PM

packages/core/src/components/dropdown/index.js
sick-sad-world: nope. i re-tested this and figured out that if we wan't to apply new options we do not need to re-render anything. Just set the value and (renderSelected) will be applied and update the option list and (renderOption) also will be invoked

sick-sad-world:
> Also want to note here that the Error state doesn't match the non-error states, is this intentional?:
> 
> * the error hint is nested within the dropdown
> * the dropdown is cut off by the message
> * the border color is not the same when open
> 
> Image 2019-12-04 at 3 53 54 PM

Nah, this is my bad - put incorrect markup. Hints in ray is adjacent to input control not inside. Will Fix this i docs asap

PR #256, opened 2019-12-02
fix(docs-deploy): update how core-js is imported

cmugla:
Was getting errors when deploying the docs-app:
Screen Shot 2019-12-02 at 2 39 07 PM

This seems to be due to updates made to `core-js`: https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md

This PR updates the import paths to be `core-js` so that it polyfills all `core-js` features

### DIFF
--- a/packages/product-react/package.json
+++ b/packages/product-react/package.json
@@ -9,7 +9,7 @@
     "dev": "yarn storybook",
     "build": "gulp scripts:es scripts:umd scripts:compiled",
     "test": "jest",
-    "storybook": "start-storybook -p 6007",
+    "storybook": "start-storybook -p 6008",
     "prebuild": "rm -rf dist/",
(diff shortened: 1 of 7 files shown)


packages/product-react/package.json
```diff
@@ -9,7 +9,7 @@
     "dev": "yarn storybook",
     "build": "gulp scripts:es scripts:umd scripts:compiled",
     "test": "jest",
-    "storybook": "start-storybook -p 6007",
+    "storybook": "start-storybook -p 6008",
```
kconst: Why the port change?

cmugla: oh both product-react and core-react used the same port `6007` - was that on purpose? I had to update when trying to run them all at the same time

PR #259, opened 2019-12-16
Fix issues related to IE11 compatibility

sick-sad-world:
Use plain `<div>` instead of `<template>` to support IE11 without polyfills
Add condition to support case when `<select>` `[selectedIndex] === -1` This case is specific for a IE browsers.
See docs: [https://deploy-preview-259--ray-docs.netlify.com/](https://deploy-preview-259--ray-docs.netlify.com/)

### DIFF
--- a/packages/core/src/components/dropdown/index.js
+++ b/packages/core/src/components/dropdown/index.js
@@ -196,7 +196,11 @@ class Dropdown {
 
   _setSelectedLabel() {
     const { renderSelected } = this.settings;
-    if (this._selectedOption > 0 || this._selectedOption === undefined) {
+    if (
+      this._selectedOption > 0 ||
+      this._selectedOption === undefined ||
+      this._value === ''
+    ) {
       this._selectedValue.innerHTML = '';
     } else {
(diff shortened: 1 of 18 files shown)


packages/core/src/components/dropdown/index.js
```diff
@@ -26,9 +26,9 @@ function tryAndPass(cond, str) {
 }
 
 function htmlToElement(html) {
-  const template = document.createElement('template');
```
wuweiweiwu: why div instead of template?

sick-sad-world: Unfortunately IE11 does not support `template`. And in `www-website` no proper polyfill provided which leading to error. Which described in this ticket https://jira.weworkers.io/browse/GROW-10169
This is solution. To our purposes it's doesn't matter what element to use for `string->dom` conversion since it does not attached to DOM tree and `"deleted"` by garbage collector after use. But you're right `template` is correct option. 
Btw as `ray` is free-to-use tool we should try to make it less dependent on other tools such as polyfills e.t.c

PR #261, opened 2020-01-10
Dropdown Enhancement: Markup re-usage

sick-sad-world:
**Bug with markup duplication observed at: https://ray.wework.com/components/dropdown/**
Root cause: Due to frequent update of Gatsby React component and possibility that DOM element (which used as Key to differ instances) may fail, so possible multi-instantiation of specific component. (Which may lead to creating exceeding markup in case of Dropdown component).
See screenshot 
Screen Shot 2020-01-10 at 4 27 27 PM

**Component enhancement as a part of solution:**
Now Dropdown will check markup that needs to be added to layout, if it already been placed (by hand, or by previous instance).
Added tests to check this specific case.

Test URL: https://deploy-preview-261--ray-docs.netlify.com/components/dropdown/

### DIFF
--- a/packages/core/src/components/dropdown/constants.js
+++ b/packages/core/src/components/dropdown/constants.js
@@ -44,6 +44,7 @@ export const SELECTORS = {
 export const bodyTpl = ({ value }) => {
   return {
     position: 'beforebegin',
+    selector: SELECTORS.body,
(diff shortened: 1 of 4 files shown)


kconst:
Thanks for the update! Looks good.

PR #262, opened 2020-03-02
Update _tag.scss

akshat-vasudev:
As per https://jira.weworkers.io/browse/GROW-10949, the text inside the tags is not vertically aligned.


packages/core/src/components/tag/_tag.scss
```diff
@@ -11,7 +11,7 @@
     text-transform: uppercase;
     font-weight: 400;
     font-size: 0.75rem;
-    line-height: 1.5;
+    line-height: 1.4;
```
akshat-vasudev, line 14: Dummy comment to retrigger build.

PR #281, opened 2021-03-30
Bump sharp dependencies

acartmell:
Fixes https://jira.weworkers.io/browse/MPWW-13422

I removed the `SHARP_DIST_BASE_URL` environment variable that tells CircleCI to pull precompiled Sharp library from s3. It seems this was set up as a workaround for some issues with Sharp at the time. Those issues have since been fixed in newer versions of our dependences, so I bumped our dependencies that rely on Sharp to the latest patch versions. CI seems to be working now.


dxueww (approved):

acartmell:
There's an issue with coveralls, need to close this PR and open a new one to resync...

